PR #35, opened 2015-03-01
Added _ref property to array objects

bs1180:
When converting a firebase snapshot into an array, this adds the individual location refs to the object in order to make subsequent manipulation easier. This is one way to address issue 33.

### DIFF
--- a/src/reactfire.js
+++ b/src/reactfire.js
@@ -126,7 +126,7 @@ var ReactFireMixin = {
         for (var key in obj) {
           if (obj.hasOwnProperty(key)) {
             var child = obj[key];
-            child['_ref'] = key;  
+            child._ref = key;  
             out.push(child);
           }
         }
--- a/src/reactfire.js
+++ b/src/reactfire.js
@@ -125,7 +125,9 @@ var ReactFireMixin = {
       else if (typeof(obj) === "object") {
         for (var key in obj) {
           if (obj.hasOwnProperty(key)) {
-            out.push(obj[key]);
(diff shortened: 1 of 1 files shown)


jwngr:
First off, thank you for submitting this PR. I really appreciate you taking a crack at this! I think while this change would work in most cases, it's not a full solution. If the array items are not object, then this won't work very well. For example, say `obj[key]` is a number. We will then do `<number>._ref = key` which doesn't really work in the context of a number. What we really need is a solution like [AngularFire's `$toJSON()` method](https://github.com/firebase/angularfire/blob/1d3260937fd5af0708d7a60795b83361dc943a29/src/utils.js#L403-L432) which handles both arrays of objects and arrays of primitives. In the case of objects, we do something very similar to what you have done. In the case of a primitive, we set the array items as objects with `$value` (i.e. the primitive value) and `$id` (the Firebase key).

I think `_key` is a more appropriate name in Firebase parlance, since `ref` is already taken. `ref` is an instance of `new Firebase()`, while `key` is the unique push ID for a Firebase array. I don't personally like AngularFire's choice of `$id` over `$key`, so I think for ReactFire, we should go with `_value` and `_key`.

If you want to take another crack at this, I'd be more than happy to review more code. Otherwise, I'll get to this eventually!

bs1180:
@jwngr Thanks for the comprehensive feedback, it's much appreciated. This was something of a drive-by PR during a Sunday afternoon project, but will have another bash at it if no-one else gets there first.

Issue #37, opened 2015-03-02
Add the Firebase key to array items


jwngr:
Currently, there is no way to get the Firebase key (aka push ID) for items bound via `bindAsArray()`. We should adopt a solution like [AngularFire's `$toJSON()` method](https://github.com/firebase/angularfire/blob/1d3260937fd5af0708d7a60795b83361dc943a29/src/utils.js#L403-L432) which handles both arrays of objects and arrays of primitives. In the case of objects, we do something very similar to what you have done. In the case of a primitive, we set the array items as objects with `$value` (i.e. the primitive value) and `$id` (the Firebase key).

jwngr:
Initial attempt: #35

matiaslarsson:
Yes please! Any progress on this?

jwngr:
Haven't had time to code this up yet. There were some attempts from the community (#35 and #40) but neither fully solved the problem. Still on my todo list.

mrienstra:
Stating the obvious here, but since this didn't immediately occur to me, perhaps it will be helpful to others:

You can use the `bindAsObject` method in place of `bindAsArray`.

Related: the [limitations of treating Firebase objects as arrays](https://www.firebase.com/docs/web/guide/understanding-data.html#section-arrays-in-firebase).

PS: If you were using `Array.prototype.map` to turn your Firebase arrays into arrays of React elements, a lazy way to instead map objects is to use [lodash's `mapValues`](https://lodash.com/docs#mapValues).

AndrewBestbier:
This was done in #40

jwngr:
Yup, this got merged in. I'll get a 0.2.0 release out with all the new goodies in the next day or two.

PR #40, opened 2015-04-26
Add `key` to the items in the array (as `$id`)

puf:
This way the key can be used to identify the items in the React output and tie updates back to the correct `child()`. See this fiddle for an example (clicking a Todo item removes it): https://jsfiddle.net/frankvanpuffelen/s27nz1hb/

### DIFF
--- a/src/reactfire.js
+++ b/src/reactfire.js
@@ -125,7 +125,9 @@ var ReactFireMixin = {
       else if (typeof(obj) === "object") {
         for (var key in obj) {
           if (obj.hasOwnProperty(key)) {
-            out.push(obj[key]);
+            var item = obj[key];
+            item['$id'] = key;
+            out.push(item);
           }
         }
       }
--- a/src/reactfire.js
+++ b/src/reactfire.js
@@ -120,12 +120,17 @@ var ReactFireMixin = {
     var out = [];
     if (obj) {
       if (this._isArray(obj)) {
-        out = obj;
+        for (var i=0; i < obj.length; i++) {
+          out.push({ $id: i, $value: obj[i] });
+        }
       }
       else if (typeof(obj) === "object") {
         for (var key in obj) {
           if (obj.hasOwnProperty(key)) {
             var item = obj[key];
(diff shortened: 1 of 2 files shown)


jwngr:
Thanks for the PR @puf! Have you seen the discussion in #37? I think we should go with a more fully-fledged solution here. As noted in the issue, I'd like to do something very similar to what we do with AngularFire. Most notably, your solution won't work for primitive values since setting `$id` on the primitive value won't actually have the effect we want.

puf:
@jwngr quick double check here....
Case 1: array of objects; if you pass in:
          firebaseRef.set({ first: { index:1 }, second: { index:2 }, third: { index:3 } });
you get:
         {"items":[{"index":1,"$id":"first"},{"index":2,"$id":"second"},{"index":3,"$id":"third"}]}

Case 2: array of primitives; you pass in:
          firebaseRef.set(["first", "second", "third"]);
You now get:
        {"items":["first","second","third"]}
You propose:
        {"items":[{$id: 0, $value: "first"}, {$id: 1, $value: "second"}, {$id: 2, $value: "third"}]}

That it? Or am I missing something? Test cases most welcome. :-)

jwngr:
Yup, that's pretty much it. That is what we do with AngularFire and I think it works out pretty well.

puf:
OK. I got that one working easily, but of course then some of the original
tests that use primitives are failing (since they don't do `.$value`). I'll
change those and update the PR.

jwngr:
Sounds good. We definitely do need to update the test suite and will also probably need to list this as a breaking change. I'm totally up for doing that though since this is strictly improved functionality.

puf:
@jwngr Update the tests and fixed a few more case. Back to you.

AndrewBestbier:
Thank you for this!

jwngr:
Thanks for this PR @puf! I made a few changes myself (renamed `$id` to `$key` and fixed issue with sparse arrays) and then merged this in. Time to crank out a few more fixes for other issues and get an updated version of ReactFire out there.

PR #45, opened 2015-07-09
Switched $key and $value to .key and .value

jwngr:


thibalte:
Hi, it's really great to be able to grab the keys of array items as I can finally edit/remove childs.

I am wondering why choosing the dot-prefixed syntax as this prevents from using dot notation to access array object props. I don't know if my limited knowledge of JS makes me miss something but wouldn't prefixing with underscores as _id or _value would be similar but permitting dot notation use?

jwngr:
Great point and I agree it's not perfect, but the underlying Firebase library will actually properly handle writes with objects that contain `.value`, so I wanted to make it as easy as possible to write back data to Firebase without having to update the object in question. For example, this will actually work with the Firebase client:

```
ref.set({
  '.value': 'foo'
});
```

While this will throw an error:

```
ref.set({
  '$value': 'foo'
});
```

So in order to write back data you got from ReactFire, all you need to do is remove the `.key` attribute, regardless of if the value is a primitive or an object.

Hopefully that all made sense!

PR #64, opened 2015-09-17
Prevent calling unbind on undefined firebaseRefs

gragland:
Currently the componentWillUnmount will attempt to unbind a ref that we've already called unbind on, resulting in an error. We need to skip any values within this.firebaseRefs that equal "undefined". Alternatively we could change unbind to delete the ref from this.firebaseRefs (it looks like it used to be that way and was changed for some reason).

### DIFF
--- a/src/reactfire.js
+++ b/src/reactfire.js
@@ -296,7 +296,7 @@
     componentWillUnmount: function() {
       for (var bindVar in this.firebaseRefs) {
         /* istanbul ignore else */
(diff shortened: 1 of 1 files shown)


gragland:
I signed the Contributor License Agreement.

jwngr:
Thanks for pointing this out and taking a crack at it. I created #69 based of your PR so that I can get out a new version. Release going out now. Thanks for the contribution!

PR #73, opened 2015-10-29
track value callback completion in firebaseLoads

joenoon:
use case: no data exists vs data not yet loaded.
scenario: this.bindAsArray(ref, "items"); without firebaseLoads
there is no way to tell why there are no items, which is a problem
if we want to show a "no items" message.  if we rely on
items.length == 0 we will see the "no items" message flicker
before the data is loaded.


joenoon:
CLA signed

PR #88, opened 2016-04-08
Update app.js

jaylandro:
Adding event.preventDefault(); to the handleSubmit method prevents the browser from submitting the form via standard mechanisms in turn engaging and demonstrating React mechanisms.

### DIFF
--- a/examples/commentsBox/js/app.js
+++ b/examples/commentsBox/js/app.js
@@ -29,6 +29,7 @@ var CommentList = React.createClass({
 
 var CommentForm = React.createClass({
   handleSubmit: function() {
+    event.preventDefault();
     var author = this.refs.author.getDOMNode().value.trim();
     var text = this.refs.text.getDOMNode().value.trim();
     this.props.onCommentSubmit({author: author, text: text});
--- a/examples/commentsBox/js/app.js
+++ b/examples/commentsBox/js/app.js
@@ -28,8 +28,8 @@ var CommentList = React.createClass({
 
 
 var CommentForm = React.createClass({
-  handleSubmit: function() {
-    event.preventDefault();
+  handleSubmit: function(e) {
+    e.preventDefault();
     var author = this.refs.author.getDOMNode().value.trim();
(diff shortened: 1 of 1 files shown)


jaylandro:
I signed it!

jwngr:
Thanks for the PR @jaylandro!

It looks like the function [already calls `return false`](https://github.com/firebase/reactfire/blob/df2e9998efa8e3fc0291c2387c8266ab61a3f409/examples/commentsBox/js/app.js#L37) which will prevent the default behavior. See [this Stack Overflow post](http://stackoverflow.com/questions/1357118/event-preventdefault-vs-return-false) for some discussion on this.

Side note: you are calling `event.preventDefault()` without including `event` as an input argument to `handleSubmit()`. So that will end up throwing one of those dreaded `"undefined is not a function"` errors.

The example should work as expected already. Are you seeing weird behavior? If so, re-open this PR and let me know what browser you are using. If you can deploy the code somewhere live which shows the issue, that would be super handy.

jaylandro:
Behavior can be witnesses here: [https://firereact-comment.firebaseapp.com/](https://firereact-comment.firebaseapp.com/) and you are so right on passing the event, I had created this PR quickly directly in github but failed to pass the event in, as I had in my other project.

Browser: Chrome/49.0.2623.110
OS: OS X 10.11.2

Thanks

jaylandro:
Documentation for React v0.12 notes that `return false` is deprecated.
[https://facebook.github.io/react/blog/2014/10/28/react-v0.12.html](https://facebook.github.io/react/blog/2014/10/28/react-v0.12.htmll)

jwngr:
Thanks for linking to the change log. It's interesting that they deprecated `return false`. I've re-opened this PR and I'll accept this PR if you do the following:

1. Add `event` as an input argument.
2. Remove the `return false;` line at the end. It doesn't do what we want anyway so let's just rip it out.

Thanks!

jaylandro:
Complete, thank you for your patience @jwngr

jwngr:
:tada: Thanks for the PR! :tada:

Issue #120, opened 2017-01-27
this.data not being cleaned up


EduardJS:
Hello! 🤙 

I'm building this messenger-like application, on top of React and Firebase, obviously, and, I've been pulling my hair out for more than 2 hours trying to find out what was causing the weird behavior. 

1. Upon login, I am listening to some rooms ( of users ), then upon room selection I start listening to messages belonging to that room. 

2. The issue arose when a user logs out, and then logs back in from another account for example.
Upon `componentWillUnmount` the references get unbinded, so that's all nice and cool.

3. So, logging back in with another user, even though there is no listener nor reference anymore, I am immediately seeing the "old" messages popping up inside the `ConversationView` component. 

After hours of trying to figure out what is happening.. I finally decided to start looking at what exactly is this mixin doing. Luckily, I quickly realised that we are storing reference data inside `this.data` but we're not deleting it when the reference gets unbinded. 

Therefore, my component's state was automatically being populated with the old reference values from within `this.data` which were saved when the previous account was logged in.

Note that it doesn't affect components which bind references inside `componentWillMount`, since those will pull new data anyway. But for the ones which are waiting for data to be binded... the "old" data ( from the previous bind -- previous logged in account ) gets sent towards the component's state once you bind any other reference.

I created a PR with a fix here #121 , which implies cleaning the unbinded variable, as it should.

samtstern:
We have officially deprecated reactfire, as we no longer believe it is the best solution for Firebase developers looking to integrate with React.

Please see the `README` for information on how to migrate to a better solution. We are working on some new efforts to better serve the React community in the future.

PR #121, opened 2017-01-27
Cleaning up reference data

EduardJS:
Described in #120

### DIFF
--- a/src/reactfire.js
+++ b/src/reactfire.js
@@ -388,6 +388,7 @@
       }
       delete this.firebaseRefs[bindVar];
       delete this.firebaseListeners[bindVar];
+      delete this.data[bindVar];
 
       // Update state
       var newState = {};


EduardJS:
I signed it!

jwngr:
Thanks for opening this issue and sending off a PR for it! I think it's a good change and I'm sorry it cost you some lost time. In addition to deleting `this.data[bindVar]`, I think we also need to call `setState()` to set the bind variable's state to `undefined`:

```
var newData = {};
newData[bindVar] = undefined;
this.setState(newData);
```

Once you make that change, I'll go ahead and merge this in.

EduardJS:
Hi @jwngr 

If I understood correctly, then I must say we already trigger `setState` on `bindVar` to `undefined` as seen here: https://github.com/firebase/reactfire/blob/master/src/reactfire.js#L392

Please let me know if that's not what you're talking about!

Cheers, and thanks for the reply!

jwngr:
Wow 😨 Not sure how I missed that... I literally wrote the same exact code as we already have in the library. Mondays...

Anyway, this looks good! Merging it in.

EduardJS:
Haha! I feel you regarding Mondays, though you missed the callback in your code  👍 

Have a good one and thanks for the merge!

PR #122, opened 2017-02-07
The inner bind is not required

ShashankaNataraj:
The inner bind is useless
create-react-apps' default webpack config was what pointed out this issue.


jwngr:
Indeed. Thank you!

PR #123, opened 2017-02-20
Implemented createAppContainer() and createAuthContainer()

jwngr:
### Description

This implements two of the higher-order components that will make up ReactFire `2.0.0`: `createAuthContainer()` and `createAppContainer()`.

`createAuthContainer()` adds a `user` prop to the provided component that is updated every time the auth state is changed. You can change the name of the prop by passing an optional second argument

`createAppContainer()`  adds a `firebaseApp` prop to the provided component. Using this in addition with `createAuthContainer()` allows you to listen for the auth state changes for the non-default Firebase app instance.

Other miscellaneous changes:
* Replaced Mocha with Jest and made use of some of Jest's snapshot testing.
* Replaced `gulp` with `npm` scripts.

### Code sample

```
const ExampleComponentWithAuth = createAuthContainer(ExampleComponent);

// Advanced use cases
const ExampleComponentWithAuthForNonDefaultApp = createAppContainer(ExampleComponentWithAuth, someApp)
const OtherComponentWithAuthForNonDefaultApp = createAppContainer(ExampleComponentWithAuth, otherApp);
```


package.json
```diff
@@ -28,42 +31,49 @@
...
+    "run-sequence": "^1.1.1"
+  },
+  "watch": {
+    "test": "{src,tests}/**/*.js"
   },
   "scripts": {
-    "test": "gulp test",
-    "travis": "gulp"
+    "test": "npm run lint ; npm run build ; jest --coverage",
+    "lint": "eslint src/ tests/",
```
tstirrat: perhaps we can use the `node_modules/.bin/` binaries here so that they do not need global installation

src/createAuthContainer.js
```diff
@@ -0,0 +1,61 @@
...
+const createAuthContainer = (WrappedComponent, userPropName = 'user') => {
+  if (!utils.isReactComponent(WrappedComponent)) {
+    throw new Error('First argument to createAuthContainer() must be a React component.');
+  } else if (typeof userPropName !== 'string' || userPropName === '') {
+    throw new Error('Second argument to createAuthContainer() must be a non-empty string representing a prop name.');
+  }
+
+  let unsubscribe;
+
+  class CreateAuthContainer extends React.Component {
```
tstirrat: nit: It'd make more sense to me if this was just called **AuthContainer**, same goes for **CreateAppContainer** -> **AppContainer**

src/utils.js
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/**
+ * Returns whether or not the provided input is a Firebase Database Reference.
+ *
+ * @param {any} input The input being tested.
+ *
+ * @return {boolean} Whether or not the provided input is a Firebase Database Reference.
+ */
+export function isDatabaseReference(input) {
```
tstirrat: doesn't appear to be used.

src/utils.js
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Returns whether or not the provided input is either a Firebase Database Reference or a Firebase
+ * Database Query.
+ *
+ * @param {any} input The input being tested.
+ *
+ * @return {boolean} Whether or not the provided input is either a Firebase Database Reference or
+ * a Firebase Database Query.
+ */
+export function isDatabaseReferenceOrQuery(input) {
```
tstirrat: same here, doesn't appear to be used.

package.json
jwngr: They actually don't need a global install. I'm not sure if this is a newer npm thing, but anything with `node_modules/.bin/` is addressable just by name in npm scripts.

src/createAuthContainer.js
jwngr: I'm conflicted and honestly not sure what the most correct thing to call this is. Let's chat about it in person tomorrow.

src/utils.js
jwngr: Will be used in the next PR. If it's cool with you, I'll just keep it in this PR.

src/utils.js
jwngr: Ditto on the above.

tstirrat:
LGTM

PR #133, opened 2018-03-16
Marking ReactFire as deprecated.

### LINKED ISSUE #132: This project is abandoned - use `react-redux-firebase` instead
At the time of posting this comment, this project:

* hasn't been updated in 11 months
* is using the [old mixin style of passing down state](https://reactjs.org/blog/2016/07/13/mixins-considered-harmful.html), which fell out of favor in 2016
* doesn't support Firestore, which was [released in October of 2017](https://firebase.googleblog.com/2017/10/introducing-cloud-firestore.html)

[react-redux-firebase](https://github.com/prescottprue/react-redux-firebase) is under active development, uses higher order component composition (the recommended replacement for mixins), and [supports Firestore like a champ](http://react-redux-firebase.com/docs/firestore.html).

You'll need to learn [redux](https://redux.js.org/) to use it, but you should probably learn it anyway. The creator of Redux, Dan Abramov, has [a fantastic free course on learning redux for beginners](https://egghead.io/courses/getting-started-with-redux).

nicolasgarnier:
Also adding a refreshed Firebase in React guide and a migration guide from ReactFire to using the JS SDK.

Rationale:

Reactfire is currently based on a deprecated/bad-practice technology (mixins).
Furthermore the current JS SDK can be used directly in React components and there are two 3rd-party Firebase libraries for React (re-base and react-redux-firebase) which are actively developed and of high quality.

fixes #132

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,80 +1,254 @@
-# ReactFire [![Build Status](https://travis-ci.org/firebase/reactfire.svg?branch=master)](https://travis-ci.org/firebase/reactfire) [![Coverage Status](https://coveralls.io/repos/firebase/reactfire/badge.svg?branch=master&service=github)](https://coveralls.io/github/firebase/reactfire?branch=master) [![GitHub version](https://badge.fury.io/gh/firebase%2Freactfire.svg)](http://badge.fury.io/gh/firebase%2Freactfire)
+# ReactFire is deprecated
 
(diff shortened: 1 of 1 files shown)


christiangenco:
Hey thank you! This is an elegant depreciation.

Issue #140, opened 2019-07-31
Incorrect type definitions in hooks?


Raicuparta:
Seems like the type definitions for `useFirestoreDoc` and `useFirestoreCollection` are incorrect:

```typescript
export declare function useFirestoreDoc<T = unknown>(ref: firestore.DocumentReference, options?: ReactFireOptions<T>): firestore.DocumentSnapshot | T;
```

[In a union, `unknown` absorbs everything](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#example-3), so if we don't pass a generic type here the return type will always be `unknown`. I tried to find a way to fix this so I could do a PR myself, but I'm afraid my TypeScript knowledge is lacking.

If I understand correctly, what we want here is to make the return type be `firestore.DocumentSnapshot` by default, or `T` if specified. I opened a PR with a way to do this but I'm really not sure if this is how it's supposed to be done.

### Version info

**React:** 16.8.6

**Firebase:** 6.0.4

**ReactFire:** 2.0.0-canary.fc2be5b

**Typescript:** 3.5.3 (on my machine) and 3.3.3 (on CodeSandox) both show the same problem

### Test case
[Test case on Code Sandbox
](https://codesandbox.io/s/reactfire-typescript-wrong-types-4wq4o)

I didn't try to make the project actually run, because the issue is just the type definitions. The issue isn't visible right away, though, because reactfire can't import from `firebase/app` on Code Sandbox for some reason. Follow these steps to see the problem:

### Steps to reproduce

1. Visit https://codesandbox.io/s/reactfire-typescript-wrong-types-4wq4o
2. Open the file `src/withoutOptions.ts`
3. Right-click `useFirestoreDoc` and select "Go To Definition""
4. Check if the `firebase/app` import is failing. If it is, replace with just `firebase`. Don't save the file, but leave it open (important).
5. Go back to `src/withoutOptions.ts` and check the type of `doc` by hovering the mouse over it.

### Expected behavior

The type should be `firestore.DocumentSnapshot` by default

### Actual behavior

The type is `unknown`

jhuleatt:
@Raicuparta thanks for this issue!

@jamesdaniels do you know what we should do here? There are two (or maybe three) ways to call these functions:

1) Just a ref:
```ts
useFirestoreDoc(ref) // returns a firestore.DocumentSnapshot
```

[...]

PR #141, opened 2019-07-31
Fix hooks return type definitions

Raicuparta:
Fix for #140


jhuleatt:
@jamesdaniels should this be the new pattern for all of the hooks that take `startWithValue` as an option?

jamesdaniels:
@jhuleatt I dig it.

jhuleatt:
@Raicuparta thanks for this PR!

PR #143, opened 2019-08-05
Add documentation for common use cases

jhuleatt:
Add a new docs page that shows common use cases


jhuleatt:
Preview here: https://github.com/FirebaseExtended/reactfire/blob/use/docs/use.md

PR #154, opened 2019-10-12
Fix SuspenseWithPerf error

### LINKED ISSUE #148: The simple-sample project does not work
I was getting errors when I was developing locally about traces, so I figured I'd pull down the demo and take a look and I'm getting a similar result:

> Trace storage-root is started before.

This happens right after you go through the authentication scheme.

jhuleatt:
fixes #148 

Fixes an edge case where SuspenseWithPerf tries to reuse a trace if a child throws another promise after resolving.


jhuleatt:
I'm going to merge this without review since it fixes a bug that blocks people from trying out reactfire.

PR #156, opened 2019-10-17
Make sure useUser works predictably in AuthCheck

### LINKED ISSUE #155: useUser initial value null within AuthCheck children
### Version info

**React:** 16.10.2

**Firebase:** 7.2.9

**ReactFire:** 2.0.0-alpha.2

### Test case

https://codesandbox.io/s/competent-mendeleev-svfjb

### Steps to reproduce

Use `AuthCheck` with fallback login page.
Login in.
In "authorized" child component, call `useUser`
Access properties of `user` without checking for null.

### Expected behavior

Since `AuthCheck` children won't render until user is logged in (not null), I'd expect all rendered descendant components under `AuthCheck` to have access to a not null user.

### Actual behavior
Because of how `useUser` instantiates a new observable for every separate usage, there will always be an initial null value. Ideally, we wouldn't have to check for this everywhere in a component tree rendered after `AuthCheck` has a user.

I don't see the use case for having an optional `startWith` parameter in `useUser` instead of defaulting to `app.auth().currentUser` and not have to pass in ourselves everywhere.

https://github.com/FirebaseExtended/reactfire/blob/99dac6c4cacef2b6fc59b53de340b935363a1e06/reactfire/auth/index.tsx#L41

jhuleatt:
fixes #155 

Due to a bug in the internals of `useObservable`, sometimes useUser would emit `null` when it was the child of an `<AuthCheck>` component, which should be impossible.

To fix this, I created a test (called _always returns a user if inside an <AuthCheck> component_) and verified that it failed in the existing ReactFire. Then, I fixed `userObservable` and verified that the test passed.

### DIFF
--- a/reactfire/auth/auth.test.tsx
+++ b/reactfire/auth/auth.test.tsx
@@ -1,18 +1,41 @@
-import { cleanup, render } from '@testing-library/react';
-import { auth } from 'firebase/app';
+import { cleanup, render, waitForElement, wait } from '@testing-library/react';
+import { auth, User } from 'firebase/app';
 import '@testing-library/jest-dom/extend-expect';
 import * as React from 'react';
-import { AuthCheck } from '.';
+import { AuthCheck, useUser } from '.';
 import { FirebaseAppProvider } from '..';
+import { Observable, Observer } from 'rxjs';
+import { act } from 'react-dom/test-utils';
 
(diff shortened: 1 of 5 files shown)


jhuleatt:
merging this since it's a bugfix and tests are passing

Issue #138, opened 2019-07-18
AuthCheck is defined as ReactNode rather than JSX.Element


waixwong:
AuthCheck is defined as ReactNode rather than JSX.Element in index.d.ts

This produces the following TS error when adding AuthCheck:
TS2605: JSX element type 'ReactNode' is not a constructor function for JSX elements.

### Version info
**React:**
16.8.6

**Firebase:**
5.10.0

**ReactFire:**
2.0.0

**Other (e.g. Node, browser, operating system) (if applicable):**
10.16

### Steps to reproduce
https://stackblitz.com/edit/react-ts-k6abtn
line 56

jhuleatt:
I originally picked `ReactNode` because that is what the [core React typings use for components like Suspense](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/752847f11e21276a22ef5703294bc689a0f3445a/types/react/index.d.ts#L347). 

It doesn't look like the typings for React use `JSX.Element` at all, so I'm a little worried about switching to that. I think this type error is more due to a [TypeScript bug](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/20544), but I'll give `JSX.Element` a shot and see if that clears up the warnings. Thanks for filing this!

pearlyplan:
Hi - is there any update on this?  Happy to do a PR with the new typing.

seanemmer:
Hi, is there any update on this?  Happy to do a PR with the new typing.

pmithunish:
Adding the following snippet helped me get rid of the `ts(2605)` error when using `AuthCheck`.
```ts
declare module "reactfire/auth" {
  function AuthCheck({
    auth,
    fallback,
    children,
    requiredClaims
  }: AuthCheckProps): JSX.Element;
}
```
Hope it helps! 😃

jhuleatt:
Sounds like the consensus is that `JSX.Element` is the way to go! @seanemmer if you have the time, a PR would be a big help! I think `<SuspenseWithPerf>` will need the same type updates.

PR #162, opened 2019-10-18
update return type for AuthCheck and SuspenseWithPerf to JSX.Element

seanemmer:
Update return type from React.ReactNode to JSX.Element

Resolves https://github.com/FirebaseExtended/reactfire/issues/138

### DIFF
--- a/reactfire/auth/index.tsx
+++ b/reactfire/auth/index.tsx
@@ -82,8 +82,8 @@ export function AuthCheck({
   });
 
   if (!user) {
-    return <>fallback</>;
+    return <>{fallback}</>;
   } else {
-    return <>children</>;
+    return <>{children}</>;
   }
 }
--- a/reactfire/auth/index.tsx
+++ b/reactfire/auth/index.tsx
@@ -54,7 +54,7 @@ export function AuthCheck({
   fallback,
   children,
   requiredClaims
-}: AuthCheckProps): React.ReactNode {
+}: AuthCheckProps): JSX.Element {
   const user = useUser<User>(auth);
 
   React.useLayoutEffect(() => {
--- a/reactfire/auth/index.tsx
+++ b/reactfire/auth/index.tsx
@@ -82,8 +82,8 @@ export function AuthCheck({
   });
 
   if (!user) {
-    return fallback;
+    return <>fallback</>;
   } else {
-    return children;
+    return <>children</>;
   }
 }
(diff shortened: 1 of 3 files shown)


seanemmer:
This passes the test suite locally - looks like 'yarn build' is failing in the CI pipeline.

jhuleatt:
`yarn build` runs the TypeScript compiler, and it looks like it is failing on a type error caused by the switch to JSX.Element:

```
auth/index.tsx:85:5 - error TS2322: Type 'ReactNode' is not assignable to type 'Element'.
  Type 'string' is not assignable to type 'Element'.
85     return fallback;
       ~~~~~~~~~~~~~~~~
auth/index.tsx:87:5 - error TS2322: Type 'ReactNode' is not assignable to type 'Element'.
  Type 'string' is not assignable to type 'Element'.
87     return children;
       ~~~~~~~~~~~~~~~~
Found 2 errors.
error Command failed with exit code 2.
```

I also got this error when running `yarn build` locally. I'll add some comments in the review that explain how to fix it.

reactfire/auth/index.tsx
```diff
@@ -54,7 +54,7 @@ export function AuthCheck({
   fallback,
   children,
   requiredClaims
-}: AuthCheckProps): React.ReactNode {
+}: AuthCheckProps): JSX.Element {
```
jhuleatt: This and the change on line 36 cause a typescript compile error because they conflict with the props declared on line 47 and 48

yarn.lock
```diff
@@ -11112,6 +11112,14 @@ react@^16.8.6:
...
     prop-types "^15.6.2"
 
+reactfire@2.0.0:
+  version "2.0.0"
+  resolved "https://registry.yarnpkg.com/reactfire/-/reactfire-2.0.0.tgz#492730ce51ee4c1b2fd5310960261090a294d4b8"
+  integrity sha512-e0IInm/0DlQDCxiIrim6Ox3x+YrPEwXBWXDzSz2WUDUz+EYOTfOSr8WpP4PKcYjLzUJBtSDHXoCqJ2boWRiMow==
+  dependencies:
+    rxfire "^3.6.3"
+    rxjs "^6.4.0"
+
```
jhuleatt: Since this isn't related to the auth types change, please remove from this PR

jhuleatt:
After playing around with the PR branch for a bit, I think this confirms that switching to JSX.Element is not the right way to go. I'm still getting warnings when I try to use a string as a child/fallback:

AuthCheck child string error

seanemmer:
I wrapped the child/fallbacks in a React fragment to handle the string scenario - this is similar to what was already done for PerfCheck

jhuleatt:
Thank you @seanemmer! Looks like it is working now 🔥

PR #164, opened 2019-10-18
[New Stuff] Auth Custom Claims

### LINKED ISSUE #161: requiredClaims for <AuthCheck> is not working.
### Version info

**React:** 16.9.0

**Firebase:** 7.2.0

**ReactFire:** 2.0.0-canary.803159c

### Test case

      <AuthCheck fallback={<Login />} requiredClaims={{ismoderator: true}}>
             <span>hello</span>
      </AuthCheck>

Am I doing anything wrong here?? Please suggest..!

### Expected behavior
It should check if user is having requiredclaims.

### Actual behavior

Throws error with undefined things.

jhuleatt:
The `AuthCheck` component had a `requiredClaims` prop, but it didn't actually work. This PR fixes that, and introduces a new Hook and Component along the way:

## `useIdTokenResult`

Wraps the Firebase JS SDK's `user.getIdTokenResult`. If you want to check custom claims, you can do this:

```jsx
const { claims } = useIdTokenResult(user, false);
```

To force a refresh, pass `true` as the second argument:

```jsx
const { claims } = useIdTokenResult(user, true);
```

## `ClaimsCheck`

Like `AuthCheck`, but only checks claims

```jsx
<ClaimsCheck
  user={user}
  fallback={<h1>Required Claims Not Met</h1>}
  requiredClaims={{admin: true}}
>
  <h1>You Have The Correct Claims</h1>
</ClaimsCheck>
```

## `AuthCheck`'s requiredClaims

If the `requiredClaims` prop is set, `AuthCheck` only renders children if a user is logged in AND they meet their required claims.

```jsx
<AuthCheck
  fallback={<h1>Not Logged In or Required Claims Not Met</h1>}
  requiredClaims={{ admin: true }}
>
  <h1>You Are Logged In and Have The Correct Claims</h1>
</AuthCheck>;
```

fixes #161

### DIFF
--- a/reactfire/auth/index.tsx
+++ b/reactfire/auth/index.tsx
@@ -2,6 +2,7 @@ import { auth, User } from 'firebase/app';
 import * as React from 'react';
 import { user } from 'rxfire/auth';
 import { useObservable, useFirebaseApp, ReactFireOptions } from '..';
+import { from } from 'rxjs';
 
 function getAuthFromContext(): auth.Auth {
   const firebaseApp = useFirebaseApp();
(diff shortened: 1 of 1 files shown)


reactfire/auth/index.tsx
```diff
@@ -42,13 +43,50 @@ export function useUser<T = unknown>(
...
 
+export interface ClaimsCheckProps {
+  user: User;
+  fallback: React.ReactNode;
+  children: React.ReactNode;
+  requiredClaims?: Object;
+}
+
+export function ClaimsCheck({ user, fallback, children, requiredClaims }) {
+  const { claims } = useIdTokenResult(user, false);
```
jhuleatt: @davideast do you think we should [`forceRefresh`](https://firebase.google.com/docs/reference/js/firebase.User.html#getidtokenresult) by default in this component?

seanaguinaga:
Does this work with a string object like:

```
  const customClaims = {
    "https://hasura.io/jwt/claims": {
      "x-hasura-default-role": "user",
      "x-hasura-allowed-roles": ["user"],
      "x-hasura-user-id": user.uid,
    },
  };
```
?

PR #165, opened 2019-10-19
Clarify versions in README

dannytatom:
### Description

Clarifying that to get 2.0, which the code examples refer to, requires installing `reactfire@next` or `reactfire@canary`. As it is, `yarn add reactfire` gives you 1.0.0 which results in a lot of errors when using the code examples and no real hint as to why.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,10 +17,12 @@ Status: Alpha
 
 ```bash
 # npm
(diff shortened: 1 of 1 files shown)


jhuleatt:
Thank you @dannytatom!

Issue #159, opened 2019-10-18
[TESTS] Realtime Database


jhuleatt:
Need to write tests for the Realtime Database hooks.

The [Firestore tests](https://github.com/FirebaseExtended/reactfire/blob/master/reactfire/firestore/firestore.test.tsx) can be used as a guide. Right now, the [`yarn test` command](https://github.com/FirebaseExtended/reactfire/blob/master/reactfire/package.json#L12) has the `--only firestore` flag set, so we'll need to change that. I think it will change to `--only firestore,database`.

Docs on using the `@firebase/test` library for Realtime Database can be found [here](https://firebase.google.com/docs/rules/unit-tests#run_local_tests).

Issue #166, opened 2019-10-20
limited ability to reactively update Firestore references/data


cdock1029:
### Version info

**React:** `16.10.2`

**Firebase:** `7.2.0`

**ReactFire:** `2.0.0-canary.803159c`

### Feature request

One of the beauties of hooks is provides concise way to "declaratively" specify how a value is calculated and when to run the recalculation given a list of dependencies.

I think the Firestore hooks could use this built in ability to recalculate their references based on provided dependency lists. At the very least the current method of updating the data based on ref path could be improved.

When using pagination to update `startAfter` for example, or flipping `orderBy` direction for sorting, or changing page size `limit`.  Or just changing the id of a `doc` when also using these other query components.. since once it's a query there is no discrimination based on `path` alone which right now is how ref updates can be triggered.

This only seems to be useful when referencing absolute collection paths. Perhaps we could switch this way of identifying a reference to something like

~~JSON.stringify(ref._query)~~
~~ref._query.toString()~~
**edit**:
```js
ref._query.canonicalId()
// https://github.com/firebase/firebase-js-sdk/blob/4ecd58ece02e08132b80ae5e50cfcd831efee762/packages/firestore/src/core/query.ts#L221
//(not sure why this method isn't public on the ref,
//seems very useful to be have a string representation of a query)
// that isn't [object Object]

```

which would allow the id of the ref to incorporate all of its query components.

A limitation of even this improvement is that only strict changes to the reference itself will refetch the data. Other non-query changes to state wouldn't, like toggling some UI data that doesn't strictly map to a path/query change.

In this case, you could have the ref be a function, allow dependency list of when to rerun this function which would return a ref:

```typescript
function useCollection(
  queryBuilder: () => firebase.firestore.Query | undefined,
  deps: any[] = [],
  // other opetional paramters
  // { get } = { get: false },
) {
  // pass that deps list to the deps parameter of some useEffect in here where a firestore 
  // subscription is created. Maybe not possible with current usage of useObservable
}
```
Allowing function to return undefined is useful because you do not always start out with everything you need to build the query (user provided ID, etc) so in that case the returned data should be undefined as well... and won't be rendered.

Usage:

```js
const page = useCollection(() => {
  let ref = undefined
  if (selectedId && !hidden) {
    ref = fs.collection('parentCollection').doc(selectedId).collection('items')
      .orderBy('name', direction).limit(pageSize)
    if (startAfter) {
      ref = ref.startAfter(startAfter)
    }
[...]

PR #167, opened 2019-10-22
Improve IDs passed to useObservable

jhuleatt:
As pointed out in #166, if two queries on the same path had different filters/sorts, ReactFire may have used one query instead of treating them as two different queries because the `observableId` wasn't specific enough. The PR fixes that issue, adds more descriptive `observableId`s overall, and adds tests to make sure we don't reintroduce this behavior in the future.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -104,10 +104,10 @@ _Throws a Promise by default_
 
 #### Parameters
 
-| Parameter   | Type                                                                                                          | Description                                                                  |
-| ----------- | ------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------- |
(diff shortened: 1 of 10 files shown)


jhuleatt:
fixes #168

jhuleatt:
Since this adds RTDB tests, it fixes #159

jhuleatt:
@davideast this PR has a bunch of code because I added RTDB emulator tests, but the important things are really:

https://github.com/FirebaseExtended/reactfire/blob/c9099aafb93b1dea76913785a6211134582652ab/reactfire/database/index.tsx#L39

and:

https://github.com/FirebaseExtended/reactfire/blob/c9099aafb93b1dea76913785a6211134582652ab/reactfire/firestore/index.tsx#L88

jhuleatt:
Thanks James! Asking to make these methods public makes a lot of sense, especially since the Android SDK already has a [`hashCode`](https://firebase.google.com/docs/reference/android/com/google/firebase/firestore/Query.html#hashCode()) method.

PR #171, opened 2019-10-25
Test and fix for error swallowing

### LINKED ISSUE #147: useObservable swallows errors
Hello.

I am using reactfire v2.
I would expect the useObservable to propagate errors which is not the case.
Below is a snippet which demonstrates the issue .

```
    try {
        useObservable(throwError(new Error('error')))
    } catch (error) {
        if (error instanceof Promise) throw error
        console.log('error catched') // this is never called
    }
```

looking at the code, `request.error` is never set.
Let me know if more detail is needed to adresse this issue.

Thanks!

jhuleatt:
fixes #147

### DIFF
--- a/reactfire/useObservable/index.ts
+++ b/reactfire/useObservable/index.ts
@@ -15,7 +15,7 @@ function suspendUntilFirst(observable$, observableId) {
       })
       .catch(err => {
         request.isComplete = true;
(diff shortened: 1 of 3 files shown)


jhuleatt:
Just going to merge this since it is pretty important and passes tests.

PR #177, opened 2019-11-05
Prevent Suspend when initialValue is null

### LINKED ISSUE #174: useObservable Suspends if startWithValue is null
If the `startWithValue` provided to `useObservable` is `null`, it causes `useObservable` to Suspend. This is caused by this line:

https://github.com/FirebaseExtended/reactfire/blob/fa4e20e68943f8f92d7a742314e91dbe668ac7d2/reactfire/useObservable/index.ts#L40-L43

We should update the code in `useObservable` to look like this:

```ts
  let initialValue;

  if (request.value !== undefined) {
    initialValue = request.value;
  } else if (startWithValue !== undefined) {
    initialValue = startWithValue;
  } else {
    initialValue = suspendUntilFirst(observable$, observableId);
  }
```

jhuleatt:
Fixes #174

### DIFF
--- a/reactfire/useObservable/index.ts
+++ b/reactfire/useObservable/index.ts
@@ -37,10 +37,15 @@ export function useObservable(
 
   const request = requestCache.getRequest(observable$, observableId);
 
-  const initialValue =
-    request.value ||
-    startWithValue ||
-    suspendUntilFirst(observable$, observableId);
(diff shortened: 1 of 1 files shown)


jhuleatt:
@prescottprue once this is merged it will unblock #172

PR #181, opened 2019-11-07
Generating CJS with rollup.js (fixes #146)

### LINKED ISSUE #146: published package is missing commonjs code
### Version info

**ReactFire:** 2.0.0

"next": "9.0.5",
"typescript": "^3.5.3"

### Test case

I just started a new project with next.js and stumbled on this new version of reactfire, I was excited to give it a try but was immediately disappointed by:
```
/home/[REDACTED]/node_modules/reactfire/index.js:1
export * from './auth';
^^^^^^

SyntaxError: Unexpected token export
```

So I opened the reactfire package and found that it doesn't contain commonjs version of source code.

### Steps to reproduce

- create a next.js based project with typescript support
- add reactfire and import it in index.tsx

### Expected behavior

Package published on npm should contain valid/runnable node.js code.

jamesdaniels:
Generating CJS using rollup, right now you need to specify the externals manually.

Fixes #146

### DIFF
--- a/reactfire/rollup.config.js
+++ b/reactfire/rollup.config.js
@@ -18,8 +18,6 @@ export default {
       'rxjs/operators',
     ],
     plugins: [
-      resolve({
-        browser: true
-      })
+      resolve()
     ]
 };
--- a/reactfire/rollup.config.js
+++ b/reactfire/rollup.config.js
@@ -1,0 +1,25 @@
+import resolve from 'rollup-plugin-node-resolve';
+
+export default {
+    input: './pub/reactfire/index.js',
+    output: {
+      file: './pub/reactfire/bundle.js',
+      format: 'cjs',
+      name: 'reactfire'
+    },
+    external: [
+      'react',
+      'firebase/app',
+      'rxfire/auth',
+      'rxfire/database',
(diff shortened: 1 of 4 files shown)


reactfire/rollup.config.js
```diff
@@ -0,0 +1,25 @@
...
+      'rxfire/auth',
+      'rxfire/database',
+      'rxfire/firestore',
+      'rxfire/storage',
+      'rxjs',
+      'rxjs/operators',
+    ],
+    plugins: [
+      resolve({
+        browser: true
```
jhuleatt: Looking [here](https://github.com/rollup/rollup-plugin-node-resolve#usage), it seems like `browser: true` is used when our package.json has a `browser` field. Our package.json doesn't have that - should we add it?

jamesdaniels: We'd want to put a browser field in our package.json if we had a UMD to link to. I had to put in here to compile as one of our dependencies needed it. I can explore further as we work to support SSR and other environments.

jamesdaniels: ... actually I don't think I need this anymore, it was dep that was getting bundled in. Missed in externals.

PR #182, opened 2019-11-09
Add a StorageImage component

### LINKED ISSUE #144: FR: Implement a <StorageImage /> component
A common Firebase pattern is to save a storage path in Firestore or the Realtime Database, and then use that to get a download URL from Cloud Storage. This is currently doable with useStorageDownloadURL like so:

```jsx
function MyComponent({path, alt}) {
  const imageURL = useStorageDownloadURL(firebaseApp.storage().ref(path));
  return (
    <img src={imageURL} alt={alt} />
  );
}
```

It would probably be helpful to just provide a component that does this for us:

```jsx
<StorageImage path="my/storage/path.jpg" alt="Yummy pasta" />
```

jhuleatt:
Fixes #144 

Takes a path to a file in a Cloud Storage bucket, triggers Suspense until it reads the downloadURL from Cloud Storage, then renders the image.

Usage:

```jsx
<StorageImage
  storagePath="path/to/my/image/in/storage/bucket.png"
  storage={useFirebaseApp().storage()} // this is optional
  // ... and any props a normal <img> component can take
/>
```

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -23,6 +23,8 @@
     - [`SuspenseWithPerf`](#SuspenseWithPerf)
   - Authentication
     - [`AuthCheck`](#AuthCheck)
+  - Cloud Storage
+    - [`StorageImage`](#StorageImage)
 - [ReactFireOptions](#ReactFireOptions)
 
 ## Providers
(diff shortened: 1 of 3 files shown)


jhuleatt:
Sample usage here: https://github.com/FirebaseExtended/reactfire/blob/2369db65f5872649923732d951dec08915051b3e/sample-simple/src/Storage.js#L67-L77

PR #206, opened 2020-01-31
[WIP] Improve observable reuse and cleanup

jhuleatt:
Main changes:

Keep multicasted observables in the request cache so that the same observable is reused for the same request
PreloadFirestoreDoc creates an observable that stays hot for 30 seconds and then cleans itself up if unused. I think this can be a pattern for the other preload functions


jhuleatt:
Closing this in favor of #210

PR #210, opened 2020-02-07
Refactor on the core observable engine

jamesdaniels:
Refactoring `useObservable` and `preloadObservable` to be more suspense friendly.

The core idea here is that our "observable" should expose a promise that resolves on first emission (which can be thrown for suspense), a synchronous means of getting the latest value (to pass into `useState`), and a shared subscription for use thereafter (for `useEffect`).

As a matter of optimization they will keep their cache around for only a present amount of time (30 seconds by default) unless they are subscribed to (`useEffect`); in which case the cache will remain until the component is unloaded. This should be enough time to prevent suspense from thrashing.

To encapsulate this behavior this we've implemented our own `Subject`.

Other changes in this PR:

* Ensure that the observable cache has to be keyed very specifically taking into account the firebase app, options, etc.
* Make sure no cache keys overlap on tests & different exports
* We're using more typescript in the tests, build them as part of the main build step since jest wasn't playing nice
* There's still some meh types kicking around here, should take as an AI
* Errors are thrown again until the cache timeout elapses, put down a TODO for now
* `FirebaseAppProvider` now throws if the provided config does not match the existing instance options, rather than simply returning the default
* `FirebaseAppProvider` now takes an `appName` property allowing you to have more than one app


reactfire/firebase-debug.log
```diff
@@ -0,0 +1,55 @@
+[debug] [2020-02-07T01:39:18.184Z] ----------------------------------------------------------------------
```
jhuleatt: Can you please add this to .gitignore? I've been meaning to do that for a while

reactfire/useObservable/useObservable.test.tsx
```diff
@@ -76,6 +77,9 @@ describe('useObservable', () => {
         }
       }
     }
+    /*
+
+    INVESTIGATE not sure what's up with the error component
```
jhuleatt: This test ensures that errors thrown by hooks can be caught by React's [Error Boundaries](https://reactjs.org/docs/error-boundaries.html)

reactfire/useObservable/useObservable.test.tsx
```diff
@@ -92,29 +96,18 @@ describe('useObservable', () => {
...
-
+*/
     spy.mockRestore();
   });
 
-  it('returns the provided startWithValue first even if the observable is ready right away', () => {
-    // This behavior is a consequense of how observables work. There is
-    // not a synchronous way to ask an observable if it has a value to emit.
-
+  it('provides the value, rather than startWithValue, when the observable is ready right away', () => {
```
jhuleatt: 👍 This behavior will save a render a decent % of the time

jhuleatt:
The CI build crashed and burned - think that's due to the new build changes? Maybe `yarn test` needs to be modified to build first?

reactfire/firebase-debug.log
jamesdaniels: Will do

reactfire/useObservable/useObservable.test.tsx
jamesdaniels: Any idea why it would be failing? Also we need to do the spy cleanup in a after block as it breaks subsequent tests if it fails.

jamesdaniels:
Yeah I didn’t update the build script, figured the right way would be to bring in your packaging work.

reactfire/useObservable/index.ts
```diff
@@ -1,89 +1,46 @@
...
-    // this will Suspend until the Promise resolves
-    initialValue = usePreloadedRequest(result);
+  const observable = preloadObservable(source, observableId);
+  if (!observable.hasValue && !startWithValue) {
+    throw observable.firstEmission;
   }
-
-  const [latestValue, setValue] = React.useState(initialValue);
-
+  const [latest, setValue] = React.useState(
```
jhuleatt: Please change to:

```ts
const [latest, setValue] = React.useState(() =>
  observable.hasValue ? observable.value : startWithValue
);
```

[`useState` will call a function that is passed to it](https://reactjs.org/docs/hooks-reference.html#lazy-initial-state), so in the case of the SDKs, they get called instead of just being used as the value.

reactfire/useObservable/index.ts
```diff
@@ -1,89 +1,46 @@
...
-        result.request.setValue(newVal);
-
-        // update state
-        setValue(newVal);
-      },
-      error => {
-        console.error('There was an error', error);
-        throw error;
+    const subscription = observable.subscribe(
+      v => setValue(v),
```
jhuleatt: Same as above, if you pass a function to setValue, [React will apparently also try to call it](https://reactjs.org/docs/hooks-reference.html#functional-updates)

So I think we need to do:

```ts
v => setValue(() => v)
```

reactfire/useObservable/useObservable.test.tsx
jhuleatt: Any chance useObservable is swallowing errors?

jamesdaniels: Needed to `catchError` and then reemit only on `.value`

reactfire/auth/index.tsx
```diff
@@ -14,9 +14,9 @@ export function preloadUser(firebaseApp: firebase.app.App) {
   return preloadAuth(firebaseApp).then(auth => {
     const result = preloadObservable(
       user(auth() as firebase.auth.Auth),
-      'auth: user'
+      `auth:preloadUser:${firebaseApp.name}`
```
jhuleatt: this should have the same observableId as `useUser` so that `useUser` can benefit from the eager fetch/warmed observable

reactfire/firestore/index.tsx
```diff
@@ -44,7 +44,7 @@ export function useFirestoreDoc<T = unknown>(
 ): T extends {} ? T : firestore.DocumentSnapshot {
   return useObservable(
     doc(ref),
-    'firestore doc: ' + ref.path,
+    `useFirestoreDoc:${ref.path}:${JSON.stringify(options)}`,
```
jhuleatt: this should have the same observableId as `preloadFirestoreDoc` above so that it can benefit from the eager fetch

reactfire/useObservable/SuspenseSubject.ts
```diff
@@ -0,0 +1,90 @@
...
+    // hasValue returns true if there's an error too
+    // so that after we resolve the promise & useObservable is called again
+    // we won't throw again
+    return this._hasValue || !!this._error;
+  }
+
+  get value(): T {
+    // TODO figure out how to reset the cache here, if I _reset() here before throwing
+    // it doesn't seem to work.
+    // As it is now, this will burn the cache entry until the timeout fires.
```
jhuleatt: I don't understand what this means

jamesdaniels: Still relevant. Say for example you (end user) hit a view you didn’t have permission for and the developer wasn’t guarding against it.

You log in, navigate to the same component again within 30 seconds, and you still see the same error screen since the cache is still kickin.

jamesdaniels: Reload the page or wait more than 30s it shows, no error.

jhuleatt: Gotcha, thanks!

reactfire/tsconfig.json
```diff
@@ -10,6 +10,5 @@
     "outDir": "pub/reactfire",
     "declaration": true,
     "skipLibCheck": true
-  },
-  "files": ["index.ts"]
+  }
```
jamesdaniels: I should .npmignore the test files now that They’re getting built in the pub dir

reactfire/firebaseApp/index.tsx
```diff
@@ -5,36 +5,54 @@ export * from './sdk';
...
         throw new Error(
           'firebase.performance not found. Did you forget to import it?'
         );
       }
-
-      // initialize Performance Monitoring
-      firebase.performance();
     }
-  }, [initPerformance, firebaseApp]);
+  }, [initPerformance, appName]);
```
jamesdaniels: undo, should be keyed on `firebaseApp`

reactfire/firebaseApp/index.tsx
```diff
@@ -5,36 +5,54 @@ export * from './sdk';
...
+  const firebaseApp: firebase.app.App =
+    props.firebaseApp ||
     React.useMemo(() => {
-      if (!firebase.apps.length) {
-        firebase.initializeApp(firebaseConfig);
+      const existingApp = firebase.apps.find(
+        app => app.name == (appName || DEFAULT_APP_NAME)
+      );
+      if (existingApp) {
+        // INVESTIGATE can we do a shallow eq check rather than JSON eq?
```
jamesdaniels: address

Issue #137, opened 2019-07-16
Implement one-time `get` for Firestore and Realtime Database


jhuleatt:
The Firebase JS SDK has a [`get`](https://firebase.google.com/docs/reference/js/firebase.firestore.Query#get) method for Firestore and Realtime Database that fetches a value once and doesn't stream updates.

We should probably implement something like a `useFirestoreDocOnce` to allow for this behavior in reactfire. Or, add a `once` flag to ReactFireOptions to accomplish the same thing.

- [x] `useFirestoreDoc[Data]Once` #211 
- [ ] `useFirestoreCollection[Data]Once`
- [ ] `useDatabaseObject[Data]Once`
- [ ] `useDatabaseList[Data]Once`

forrest-keller:
I think we should use `getFirestoreDoc`, as used in the official Firebase APK

Issue #202, opened 2020-01-21
Handle undefined gracefully in hooks?


davepwsmith:
I'm not sure if this is a deliberate choice to force me to think harder about breaking my app down into teeny-tiny pieces, but for discussion anyway: 

The nature of hooks is that they can't be called conditionally. But if I am using a field from another document which should contain a document reference to feed into `useFirebaseDocData` or some-such, then by the nature of a schemaless database that reference might not actually be there. This means that in some cases, it is hard to guarantee that what I am trying to hand to the hook isn't `undefined`. 

`react-firebase-hooks` handles this by (as far as I can tell) returning `undefined` when you feed it `undefined` - which means that you can handle this lack of data from within the component. As it stands, reactfire's hooks just throw their arms up in the air (at least in typescript-land) which means that instead I need to split out every component that relies on data from more than one place into its own thing, and then render the whole component conditionally based on the availability of the data. It may seem like the right approach, but in practice it seems to mean lots of small, not at all re-usable components end up getting made.

Your thoughts?

jhuleatt:
Hi @davepwsmith!

It looks like you caught a bug in ReactFire. It turns out that the library that underpins ReactFire, RxFire, doesn't handle the `undefined` case correctly. I've filed an issue with more detail [here](https://github.com/firebase/firebase-js-sdk/issues/2561). In cases where it should return `null` or `undefined`, it's actually returning `{}`.

davepwsmith:
So - that definitely looks like a bug in rxfire... _but_ I'm not sure still that hooks shouldn't handle this a little differently? 

When you use the firebase JS SDK without hooks (or even in your own custom hooks) because you aren't using hooks, you can always wrap things in conditionals. So for example if I want to get a document based on the users UID I can make that fetch conditional on the user object existing - pesudocode: 

```typescript
const user = firebase.auth().currentUser
if(user && user.uid) {
    return {...user, ...getFirestoreDataByUid(user.uid)}
} else { return user }
```

But because you can't call hooks conditionally, you can't do a similar thing with react hooks, so if I want to get a user (`useUser()`) and then fetch their profile (`useFirestoreDocData()`) I might end up actually _passing_ undefined into the hook, resulting in various contortions e.g.:

```typescript
// This could be null, in which case user.uid is undefined 
const user = useUser()

// - or in the case of this hook even 'unknown', so we need a typeguard

function isUser(user: any): user is firebase.User {
  return user.uid && user.uid !== null;
}

[...]

PR #211, opened 2020-02-09
Add hooks to get firestore document data once

davepwsmith:
Had some spare time and thought I would (try to) contribute instead of just moaning on the issues :-) 

This is part of the way to #137 

Not sure if this is the most efficient way to do this, but it fits in well with the general pattern. 

Presume you will need me to document it (although it will mostly be a copy and paste I should think!!) - I'll do that next week.


jhuleatt:
@davepwsmith thank you for reviving #137 and for the thorough PR with tests and sample code!

My main question is whether we instead want to handle this in `ReactFireOptions` with a `once` param or similar instead of writing out dedicated `*Once` functions. I'm meeting with @davideast and @jamesdaniels later today, so I'll run this by them too. Any thoughts on the tradeoffs?

davepwsmith:
I don't really see that there is a meaningful difference between those two approaches, from the point of view of a user of the library (of which I am one!) - I would point out that the names of the hooks are getting pretty long! I guess it _could_ be one less thing to import, but I haven't often found that I am using multiple hooks in one component where I can avoid it - not least because #202 precludes passing data between them easily.

I also wanted to take a look at doing the same for collections, but somehow I wasn't sure it seemed as simple. I was using the sample code partly as a way of working out when the observables were returning (the hooks seem to return a value more than once which is interesting... I guess that is the component getting re-rendered, I didn't look into it too hard) and when I tried the same approach with collections on the face of it, it seemed to return the observable every time the collection was modified - still with the first value, but nonetheless, not "Once". I might take another look at that and check that it isn't the sample code itself that is causing that though. 

Let me know any further thoughts.

reactfire/firestore/index.tsx
```diff
@@ -49,6 +50,23 @@ export function useFirestoreDoc<T = unknown>(
...
+ *
+ * @param ref - Reference to the document you want to get
+ * @param options
+ */
+export function useFirestoreDocOnce<T = unknown>(
+  ref: firestore.DocumentReference,
+  options?: ReactFireOptions<T>
+): T extends {} ? T : firestore.DocumentSnapshot {
+  return useObservable(
+    doc(ref).pipe(first()),
```
jhuleatt: todo: @davideast mentioned that there may be a problem with `first` never unsubscribing if the document doesn't exist. But not 100% sure

davepwsmith: I looked this up and I _think_ that first (as opposed to take(1)) throws an error if the stream completes with no values - https://swalsh.org/blog/rxjs-first-vs-take1 - but I'm a little new to observables, so I could be wrong. Might be something that I could write into a test!

davepwsmith: Just added another test (for useFirestoreDocOnce, since rxfire returns empty objects for docData) using a non-existent doc reference... seeeems to work as intended, but I am not sure how I could get inside the hooks and work out if the observable is actually unsubscribed?

See commit below.

jhuleatt: Makes sense to me! We were trying to remember whether `take(1)` or `first` was the problematic one. Thanks for following up!

reactfire/firestore/index.tsx
```diff
@@ -49,6 +50,23 @@ export function useFirestoreDoc<T = unknown>(
...
+    'firestore doconce: ' + ref.path,
```
jhuleatt: To match the incoming #210, please change this to:

```ts
`useFirestoreDocOnce:${ref.path}:${JSON.stringify(options)}`
```

reactfire/firestore/index.tsx
```diff
@@ -66,6 +84,23 @@ export function useFirestoreDocData<T = unknown>(
...
+ * @param ref - Reference to the document you want to get
+ * @param options
+ */
+export function useFirestoreDocDataOnce<T = unknown>(
+  ref: firestore.DocumentReference,
+  options?: ReactFireOptions<T>
+): T {
+  return useObservable(
+    docData(ref, checkIdField(options)).pipe(first()),
+    'firestore docdataonce: ' + ref.path,
```
jhuleatt: To match the incoming #210, please change this to:

```ts
`useFirestoreDocDataOnce:${ref.path}:${JSON.stringify(options)}`
```

reactfire/firestore/index.tsx
```diff
@@ -49,6 +50,23 @@ export function useFirestoreDoc<T = unknown>(
...
+ * @param ref - Reference to the document you want to get
+ * @param options
+ */
+export function useFirestoreDocOnce<T = unknown>(
+  ref: firestore.DocumentReference,
+  options?: ReactFireOptions<T>
+): T extends {} ? T : firestore.DocumentSnapshot {
+  return useObservable(
+    doc(ref).pipe(first()),
+    `useFirestoreDocOnce:${ref.path}:${JSON.stringify(options)}`,
```
jamesdaniels: Sorry #210 is a bit of a moving target. Updated the pattern again, let's switch this to `firestore:docOnce:${ref.firestore.app.name}:${ref.path}` and the other one to `firestore:docDataOnce:${ref.firestore.app.name}:${ref.path}:idField=${idField}`. That way we're ensuring we're caching the observable correctly. I'm going to write a util function that does this automatically in the future.

davepwsmith: Would you like me to update all the functions as above, or just the ones I have added?

davepwsmith: Also, presumably ${idField} may not exist?

jamesdaniels: Yeah that won't be a problem, I just want to make sure that if the developer doesn't get unexpected results from it using a prior cached observable when they change `idField`.

jamesdaniels:
Thanks for the contribution @davepwsmith! Look for it in the next release.

Issue #214, opened 2020-02-16
Unable to resolve "./index.esm-e07042b9.js" from "node_modules/reactfire/cjs/index.js"


josermorales:
### Version info

**React:16.9.0

**Firebase:7.8.5

**ReactFire:2.0.0-rc.0

**Other (e.g. Node, browser, operating system) (if applicable):**

OS: Mac OS 10.15.3

### Steps to reproduce

1. New react project (react native expo in my case)
2. yarn add reactfire
3. yarn add firebase
4. Wrap app in FB provider with config
5. Start server
6. Unable to resolve "./index.esm-e07042b9.js" from "node_modules/reactfire/cjs/index.js"

### Expected behavior

react hooks available

### Actual behavior

Unable to resolve "./index.esm-e07042b9.js" from "node_modules/reactfire/cjs/index.js"

jamesdaniels:
I have a good handle on what this might be, will get it fixed in rc.1. Thanks for filing the issue!

jamesdaniels:
Should be addressed in `2.0.0-rc.1` which we just cut to NPM, if you have any other problems just open another issue.

Issue #207, opened 2020-01-31
Remove `initPerformance` from `FirebaseAppProvider`


jhuleatt:
Since #205 introduces a way to initialize the different Firebase SDKs, we probably don't need to include `initPerformance` in `FirebaseAppProvider` any more

PR #215, opened 2020-02-19
Refactoring use/preload x

jamesdaniels:
* Drop `initPerformance` #207
* Cleaning up the use/preload types
* Use my good friend `Proxy` to allow access to both the `serviceProps` and the initialized instance without double calling (e.g, `useAuth()()`)
* More checks against double initializing
* Added binding to the initialize call #213 
* `SuspenseWithPerf` now lazy loads performance monitoring and uses the user timing API
* Fixed the CJS build with a better external test #214

## API change:

`use*` functions no longer provide an instance factory, you no longer need to call them again to make calls against the Firebase product.

Developers targeting older browsers will need a Proxy polyfill.

E.g, what was `useAuth()` is now `useAuth` and `useAuth()()` is now `useAuth()`

This allows our API to better match the vanilla JS SDK.

```ts
const Counter = props => {
  const firestore = useFirestore; // was useFirestore()
  const serverIncrement = firestore.FieldValue.increment;
  const ref = firestore().doc('count/counter');
  ...
};
```

```ts
const StaticValue = props => {
  const firestore = useFirestore(); // was useFirestore()()
  const ref = firestore.doc('count/counter');
  ...
};
```

### DIFF
--- a/reactfire/firebaseApp/sdk.tsx
+++ b/reactfire/firebaseApp/sdk.tsx
@@ -71,11 +71,12 @@ function proxyComponent(
   return new Proxy(componentFn, {
     get: (target, p) => target()[p],
     apply: (target, _this, args) => {
+      const component = target().bind(_this);
       // If they don't pass an app, assume the app in context rather than [DEFAULT]
       if (!args[0]) {
         args[0] = contextualApp;
       }
-      return target().bind(_this)(...args);
+      return component(...args);
     }
   }) as any;
 }
--- a/reactfire/firebaseApp/sdk.tsx
+++ b/reactfire/firebaseApp/sdk.tsx
@@ -1,219 +1,117 @@
-import { useFirebaseApp } from '..';
-import { useObservable } from '../useObservable';
-import { from } from 'rxjs';
 import { DEFAULT_APP_NAME } from '../index';
 
+type ComponentName = "analytics" | "auth" | "database" | "firestore" | "functions" | "messaging" | "performance" | "remoteConfig" | "storage";
+type ValueOf<T> = T[keyof T];
 type App = import('firebase/app').app.App;
-type RemoteConfig = Omit<typeof import('firebase/app').remoteConfig, "*"> & import('firebase/app').remoteConfig.RemoteConfig;
-type Storage = Omit<typeof import('firebase/app').storage, "*"> & import('firebase/app').storage.Storage;
-type Firestore = Omit<typeof import('firebase/app').firestore, "*"> & import('firebase/app').firestore.Firestore;
-type Performance = Omit<typeof import('firebase/app').performance, "*"> & import('firebase/app').performance.Performance;
-type Messaging = Omit<typeof import('firebase/app').messaging, "*"> & import('firebase/app').messaging.Messaging;
-type Functions = Omit<typeof import('firebase/app').functions, "*"> & import('firebase/app').functions.Functions;
-type Database = Omit<typeof import('firebase/app').database, "*"> & import('firebase/app').database.Database;
-type Auth = Omit<typeof import('firebase/app').auth, "*"> & import('firebase/app').auth.Auth;
-type Analytics = Omit<typeof import('firebase/app').analytics, "*"> & import('firebase/app').analytics.Analytics;
+type FirebaseInstanceFactory = ValueOf<Pick<App, ComponentName>>;
+type FirebaseNamespaceComponent = ValueOf<Pick<typeof import('firebase/app'), ComponentName>>;
 
-type FirebaseNamespace =
-  | Analytics
-  | Auth
-  | Database
-  | Firestore
-  | Functions
-  | Messaging
-  | Performance
-  | RemoteConfig
-  | Storage;
-
-const enum SDK {
-  ANALYTICS = 'analytics',
-  AUTH = 'auth',
-  DATABASE = 'database',
-  FIRESTORE = 'firestore',
-  FUNCTIONS = 'functions',
-  MESSAGING = 'messaging',
-  PERFORMANCE = 'performance',
-  REMOTE_CONFIG = 'remoteConfig',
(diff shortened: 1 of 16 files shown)


reactfire/firebaseApp/sdk.tsx
```diff
@@ -36,21 +39,23 @@ enum SDK {
...
-
-  let sdkPromise: Promise<FirebaseSDK>;
-
   if (firebaseApp[sdk]) {
     // Don't apply settings here. Only needed for lazy loaded SDKs.
     // If not lazy loaded, user can provide settings as normal
-    sdkPromise = Promise.resolve(firebaseApp[sdk]);
+    if (settingsCallback) { console.warn(`${sdk} was already initialized on ${firebaseApp.name == DEFAULT_APP_NAME ? 'the default app' : firebaseApp.name }, ignoring settingsCallback`) }
+    // TODO cleanup types, this is an internal API
+    const serviceProps = (firebaseApp as any).container?.providers?.get(sdk)?.component?.serviceProps ?? {};
```
jamesdaniels: DRY this up

jamesdaniels:
Need to fix the perf-suspense integration

reactfire/firebaseApp/sdk.tsx
```diff
@@ -1,28 +1,31 @@
...
-  | (() => firebase.messaging.Messaging)
-  | (() => firebase.performance.Performance)
-  | (() => firebase.remoteConfig.RemoteConfig)
-  | ((url?: string) => firebase.storage.Storage);
-
-enum SDK {
+import { DEFAULT_APP_NAME } from '../index';
+
+type App = import('firebase/app').app.App;
+type RemoteConfig = Omit<typeof import('firebase/app').remoteConfig, "*"> & import('firebase/app').remoteConfig.RemoteConfig;
```
jamesdaniels: This seems to omit the constructor, probably want to clean up.

reactfire/firebaseApp/sdk.tsx
```diff
@@ -119,100 +123,97 @@ function useSDK(sdk: SDK, firebaseApp?: firebase.app.App) {
 }
 
 export function preloadAuth(
-  firebaseApp: firebase.app.App,
-  settingsCallback?: (auth: () => Auth) => void
```
jamesdaniels: Can I infer from tyepscript?

jamesdaniels: Also should revert regardless, since I can't guarantee side-effects haven't been stripped.

reactfire/firebaseApp/sdk.tsx
```diff
@@ -1,28 +1,31 @@
...
-  | (() => firebase.performance.Performance)
-  | (() => firebase.remoteConfig.RemoteConfig)
-  | ((url?: string) => firebase.storage.Storage);
-
-enum SDK {
+import { DEFAULT_APP_NAME } from '../index';
+
+type App = import('firebase/app').app.App;
+type RemoteConfig = Omit<typeof import('firebase/app').remoteConfig, "*"> & import('firebase/app').remoteConfig.RemoteConfig;
+type Storage = Omit<typeof import('firebase/app').storage, "*"> & import('firebase/app').storage.Storage;
```
jhuleatt: What does this `Omit` syntax do?

jamesdaniels: It’s strange but this strips the constructor from the type, I’m going to clean this up more before pulling this in. I think I do want to go with the pattern that `useAuth == firebase.auth` and `useAuth() == firebase.auth()` to better mirror the JS SDK. Also chatting with @Feiyang1 to figure out why the props aren’t present on the namespace, so hopefully I don’t have to lean on Proxy & the internal API there.

davepwsmith: Unsolicited opinion warning but: to my mind, this makes a lot more sense - the closer the results returned from hooks like useFirestore() are to the vanilla firebase sdk the better, and it can take advantage of all the types.

jamesdaniels: Addressed.

jamesdaniels:
Made some more bigger changes here, going to fix the tests and update my PR comment.

reactfire/performance/index.tsx
```diff
@@ -9,32 +9,13 @@ export interface SuspensePerfProps {
...
-}
-
 export function SuspenseWithPerf({
   children,
   traceId,
   fallback,
   firePerf
 }: SuspensePerfProps): JSX.Element {
-  firePerf = firePerf || getPerfFromContext();
+  firePerf = firePerf || usePerformance();
```
jamesdaniels: Preload perf and use the user timing API in the meantime

reactfire/firebaseApp/sdk.tsx
```diff
@@ -1,218 +1,117 @@
...
+function proxyComponent(componentName: "analytics"   ): typeof import('firebase/app').analytics;
+function proxyComponent(componentName: "database"    ): typeof import('firebase/app').database;
+function proxyComponent(componentName: "firestore"   ): typeof import('firebase/app').firestore;
+function proxyComponent(componentName: "functions"   ): typeof import('firebase/app').functions;
+function proxyComponent(componentName: "messaging"   ): typeof import('firebase/app').messaging;
+function proxyComponent(componentName: "performance" ): typeof import('firebase/app').performance;
+function proxyComponent(componentName: "remoteConfig"): typeof import('firebase/app').remoteConfig;
+function proxyComponent(componentName: "storage"     ): typeof import('firebase/app').storage;
+function proxyComponent(componentName: ComponentName): FirebaseNamespaceComponent {
+  return new Proxy(() => useFirebaseComponent(componentName), {
```
jamesdaniels: Downside, I can't seem to use react context within the proxy... keep investigating else I won't be able to drop the `()()` style.

PR #218, opened 2020-02-24
Using our subject/cache to fix the init race I introduced

jamesdaniels:
In my prior refactor I introduced two regressions:

1. `useFirestore()`, etc. would not throw a Promise if the init callback from `preloadFirestore` was still awaiting. This is important for `firestore.enablePersistence()` etc. to work correctly.
1. The init callback would not run if you called preload on a second app, since it was checking merely for existence on the namespace. We'll lean on the JS SDK's errors for double initialization.
1. It turns out that since we're side-effect free, webpack "might" copy our module, so our cache needs to be "global". I'm now attaching our preload cache to `globalThis`.

### DIFF
--- a/reactfire/firebaseApp/index.tsx
+++ b/reactfire/firebaseApp/index.tsx
@@ -6,7 +6,7 @@ export * from './sdk';
 type FirebaseAppContextValue = firebase.app.App;
 
 // INVESTIGATE I don't like magic strings, can we have export this in js-sdk?
-export const DEFAULT_APP_NAME = '[DEFAULT]';
+const DEFAULT_APP_NAME = '[DEFAULT]';
 
 const FirebaseAppContext = React.createContext<
   FirebaseAppContextValue | undefined
--- a/sample/package.json
+++ b/sample/package.json
@@ -5,6 +5,7 @@
   "dependencies": {
     "@types/jest": "^24.9.0",
     "firebase": "^7.9.1",
+    "js-levenshtein": "^1.1.6",
     "react": "0.0.0-experimental-5de5b6150",
     "react-dom": "0.0.0-experimental-5de5b6150",
     "react-firebaseui": "^4.0.0",
--- a/reactfire/firebaseApp/sdk.tsx
+++ b/reactfire/firebaseApp/sdk.tsx
@@ -1,6 +1,7 @@
-import { DEFAULT_APP_NAME } from '../index';
 import { useFirebaseApp } from '.';
 import * as firebase from 'firebase/app';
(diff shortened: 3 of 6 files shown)


reactfire/firebaseApp/index.tsx
```diff
@@ -6,7 +6,7 @@ export * from './sdk';
 type FirebaseAppContextValue = firebase.app.App;
 
 // INVESTIGATE I don't like magic strings, can we have export this in js-sdk?
-export const DEFAULT_APP_NAME = '[DEFAULT]';
+const DEFAULT_APP_NAME = '[DEFAULT]';
```
jhuleatt: looks like this might not be needed any more?

sample/package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "@types/jest": "^24.9.0",
     "firebase": "^7.9.1",
+    "js-levenshtein": "^1.1.6",
```
jhuleatt: why is this needed?

jhuleatt:
Want to take a shot at writing a firestore test to confirm that preloadFirestore safely sets the settings before any other calls are made?

reactfire/firebaseApp/index.tsx
jamesdaniels: I'm still using within this module (`FirebaseAppProvider`) no longer using in `sdk.ts`, thus dropping the export.

sample/package.json
jamesdaniels: TBH, not sure, new peer dep somewhere in the tree. The demo app wouldn't compile without it.

jamesdaniels:
@jhuleatt def. Going to write some defensive tests here to make sure we don't regress again. Also will permutate over default vs. contextual and multiple apps.

sample/package.json
jamesdaniels:
```
Failed to compile.

../node_modules/react-dev-utils/webpackHotDevClient.js
Error: [BABEL] /Users/james/Code/Firebase/reactfire/node_modules/react-dev-utils/webpackHotDevClient.js: Cannot find module 'js-levenshtein' (While processing: "/Users/james/Code/Firebase/reactfire/node_modules/babel-preset-react-app/dependencies.js")
```

reactfire/firebaseApp/index.tsx
jhuleatt: 👍

sample/package.json
jhuleatt: Huh. I bumped the react version on master - maybe you need a fresh `yarn install`?

PR #220, opened 2020-02-26
Fix useObservable edge case

jhuleatt:
If a component changed the observable it passed to `useObservable`, `useObservable` wouldn't emit the new observable's first value.

This happened for 2 reasons:

1. React didn't reset `useObservable`'s `latest` value (by calling the function passed to `useState`). This is where `useObservable` usually gets the first value.
1. Because `useObservable` subscribed to the new observable in a `useEffect` call, that subscribe happens async and often *after* the observable emits its first value.

Changing `share` to `shareReplay` in `SuspenseSubject` causes the new observable to emit when the subscribe happens in `useObservable`, fixing this bug.

### DIFF
--- a/reactfire/useObservable/SuspenseSubject.ts
+++ b/reactfire/useObservable/SuspenseSubject.ts
@@ -1,5 +1,5 @@
-import { Observable, Subject, Subscription, Subscriber, empty } from 'rxjs';
-import { tap, share, catchError } from 'rxjs/operators';
+import { empty, Observable, Subject, Subscriber, Subscription } from 'rxjs';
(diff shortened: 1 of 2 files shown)


reactfire/useObservable/SuspenseSubject.ts
```diff
@@ -30,7 +28,7 @@ export class SuspenseSubject<T> extends Subject<T> {
         }
       ),
       catchError(() => empty()),
-      share()
+      shareReplay(1)
```
jhuleatt: @jamesdaniels this is the important line

Issue #221, opened 2020-03-08
Error evaluating 'query._query.canonicalId' when using useFirestoreCollectionData


jylin:
### Version info

**React:** 16.13.0

**Firebase:** 7.10.0

**ReactFire:** 2.0.1

**Other (e.g. Node, browser, operating system) (if applicable):**

### Test case

The sample in the reactfire codebase seems to throw this exception without any changes.

### Steps to reproduce

Run the sample app included in this codebase.

### Expected behavior

No crash

### Actual behavior

```
TypeError: undefined is not an object (evaluating 'query._query.canonicalId')

getHashFromFirestoreQuery
node_modules/reactfire/firestore/index.js:33
  30 |     return useObservable(fromCollectionRef(query), queryId, checkStartWithValue(options));
  31 | }
  32 | function getHashFromFirestoreQuery(query) {
> 33 |     return query._query.canonicalId();
  34 | }
  35 | export function useFirestoreCollectionData(query, options) {
  36 |     var idField = checkIdField(options);
```

jylin:
I reverted to firebase version 7.5.0, and that seems to fix it, so it's probably due to a change to firebase rather than in reactfire. Didn't try bisecting to see which exact version/change causes the issue though.

16oh4:
I'm having the same issue with Firebase 7.10.0 and the rest of these deps `"@material-ui/core": "^4.9.5",
    "@testing-library/jest-dom": "^4.2.4",
    "@testing-library/react": "^9.3.2",
    "@testing-library/user-event": "^7.1.2",
    "firebase": "^7.10.0",
    "react": "^0.0.0-experimental-235a6c4af",
    "react-dom": "^0.0.0-experimental-235a6c4af",
    "react-scripts": "3.4.0",
    "reactfire": "^2.0.1"`

pablogravielseo:
Same problem here, I was using version "7.9.3", now I've installed the version "v7.8.2" and is working.

jamesdaniels:
I believe this is due to Firebase 7.9 mangling the private API that we rely on in CJS builds... thoughts @schmidt-sebastian?

schmidt-sebastian:
Talked to @jamesdaniels, who will add an externs file for the private APIs that are used here and add it here: https://github.com/firebase/firebase-js-sdk/blob/master9/packages/firestore/externs.json

jamesdaniels:
[...]

PR #224, opened 2020-03-11
Don't use JS SDK private APIs.

jamesdaniels:
Don't mind noise due to the recent linting changes.

No longer use the undocumented `queryIdentifier` method and `cannonicalId` property in RTDB and Firestore queries respectively. A) they're not stable B) `cannonicalId` doesn't give us the properties we expect and C) `._query.cannonicalId` is getting mangled in the minified Firestore builds #221. Instead we should be using the exposed `isEqual` method to compare.

Now there is a break here, I needed to store these query caches globally & it's very important they work correctly across platforms. As such I decided to drop the flawed inline polyfill I already had for `globalThis`. I've included a polyfill in our test suite but I'm not bundling it. 

We should communicate polyfills a developer might need in the docs. I will take a stab at that in this PR.

### DIFF
--- a/reactfire/database/index.tsx
+++ b/reactfire/database/index.tsx
@@ -1,14 +1,31 @@
 import { database } from 'firebase/app';
 import { list, object, QueryChange, listVal } from 'rxfire/database';
-import {
-  ReactFireOptions,
-  useObservable,
-  checkIdField,
-  checkStartWithValue
-} from '..';
+import { ReactFireOptions, useObservable, checkIdField, checkStartWithValue } from '..';
 
 import { Observable } from 'rxjs';
 import { map } from 'rxjs/operators';
+
+const QUERY_UNIQUE_IDS = '_reactFireDatabaseQueryUniqueIds';
+
+// Since we're side-effect free, we need to ensure our observableId cache is global
+const cachedUniqueIds: Map<database.Query, string> = globalThis[QUERY_UNIQUE_IDS] || new Map();
+
+if (!globalThis[QUERY_UNIQUE_IDS]) {
+  globalThis[QUERY_UNIQUE_IDS] = cachedUniqueIds;
+}
+
+function getUnqiueIdForDatabaseQuery(query: database.Query) {
+  for (const [cachedQuery, cachedUniqueId] of cachedUniqueIds.entries()) {
+    if (cachedQuery.isEqual(query)) {
+      return cachedUniqueId;
+    }
+  }
+  const uniqueId = Math.random()
+    .toString(36)
+    .split('.')[1];
+  cachedUniqueIds.set(query, uniqueId);
+  return uniqueId;
+}
 
 /**
  * Subscribe to a Realtime Database object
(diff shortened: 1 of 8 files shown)


reactfire/database/index.tsx
```diff
@@ -1,30 +1,40 @@
...
+  globalThis[QUERY_UNIQUE_IDS] = cachedUniqueIds;
+}
+
+function getUnqiueIdForDatabaseQuery(query: database.Query) {
+  for (const [cachedQuery, cachedUniqueId] of cachedUniqueIds.entries()) {
+    if (cachedQuery.isEqual(query)) {
+      return cachedUniqueId;
+    }
+  }
+  const uniqueId = Math.random()
```
jamesdaniels: I could potentially flip the key/value around to check for collision...

jamesdaniels:
FYI @schmidt-sebastian

reactfire/database/index.tsx
```diff
@@ -1,30 +1,43 @@
...
+      return cachedUniqueId;
+    }
+  }
+  let uniqueId: string;
+  do {
+    uniqueId = Math.random()
+      .toString(36)
+      .split('.')[1];
+  } while (cachedUniqueIds.has(uniqueId));
+  cachedUniqueIds.set(uniqueId, query);
```
schmidt-sebastian: Couldn't you just use an array for `cachedUniqueIds ` and use the array index as the unique ID?

jamesdaniels: Fair! I already had this code kicking around, so I reused it. Could just be as simple as an array thanks.

jamesdaniels: Thanks again for helping get this sorted. Addressed.

reactfire/database/index.tsx
```diff
@@ -1,30 +1,35 @@
...
+const CACHED_QUERIES = '_reactFireDatabaseCachedQueries';
+
+// Since we're side-effect free, we need to ensure our observableId cache is global
+const cachedQueries: Array<database.Query> = globalThis[CACHED_QUERIES] || [];
+
+if (!globalThis[CACHED_QUERIES]) {
+  globalThis[CACHED_QUERIES] = cachedQueries;
+}
+
+function getUnqiueIdForDatabaseQuery(query: database.Query) {
```
jhuleatt: nit: getUnqiue => getUnique

reactfire/firestore/index.tsx
```diff
@@ -1,40 +1,41 @@
...
+const CACHED_QUERIES = '_reactFireFirestoreQueryCache';
+
+// Since we're side-effect free, we need to ensure our observableId cache is global
+const cachedQueries: Array<firestore.Query> = globalThis[CACHED_QUERIES] || [];
+
+if (!globalThis[CACHED_QUERIES]) {
+  globalThis[CACHED_QUERIES] = cachedQueries;
+}
+
+function getUnqiueIdForFirestoreQuery(query: firestore.Query) {
```
jhuleatt: nit: getUnqie => getUnique

jamesdaniels: Thx will address

PR #250, opened 2020-04-14
Use observable value directly in useObservable

davidhewitt:
Today I came across an interesting problem with `useObservable`.

Imagine the following situation.

- User creates a custom observable using a memoized `combineLatest`. I pass this observable to `useObservable`.
- Later on, app parameters change and a new memoized observable is created.

Now, there's an interesting bug going on where for _one render_ between the previous and new state I see an intermediate result from `useObservable` where it's using the new observable but the output value is the last value seen from the old observable.

The cause seems to be the use of `setState` on line 36: when the observable is changed, the state still contains the value of the old observable, and won't update until after the render is completed and the `useEffect` subscription runs.

So there's one render where the invariants in my app are broken and I see exceptions / warnings.

A fix seems to be to return `observable.value` directly instead of the stored state, because that's always up to date. The `useState` is still necessary to cause updates when the output changes, but we ignore the state value itself.

### DIFF
--- a/reactfire/useObservable/index.ts
+++ b/reactfire/useObservable/index.ts
@@ -33,15 +33,15 @@ export function useObservable<T>(source: Observable<T | any>, observableId: stri
   if (!observable.hasValue && !startWithValue) {
     throw observable.firstEmission;
   }
-  const [latest, setValue] = React.useState(() => (observable.hasValue ? observable.value : startWithValue));
+  const [, update] = React.useState(() => (observable.hasValue ? observable.value : startWithValue));
   React.useEffect(() => {
     const subscription = observable.subscribe(
-      v => setValue(() => v),
+      v => update(() => v),
       e => {
         throw e;
       }
     );
     return () => subscription.unsubscribe();
   }, deps);
-  return latest;
+  return observable.value;
 }


davidhewitt:
(I see that CI fails because this change is incompatible with `startVal`. That's unsuprising, and I'd be happy to fix this up, but would like some input first on whether this solution is the right direction.)

davidhewitt:
An alternative solution could be to instead store the `[lastest, observableId]` tuple in the state and return `startVal` (or undefined) if the observable id in the state doesn't match the current observable id.

(Or perhaps a variant of the tuple which uses the dependency list? I'm not sure how easy it is for us to compare two dependency lists...)

jhuleatt:
Hey @davidhewitt, thanks for the PR, and apologies for not responding earlier. I believe this bug was addressed as part of https://github.com/FirebaseExtended/reactfire/pull/255 - now, the `useEffect` hook is passed the observable in its `deps` array, instead of the `observableId`. This means that swapping out the observable without changing `observableId` will force a re-render in cases where it wouldn't have previously.

However, if you can create a test that catches the in-between state with the latest ReactFire code, I'd be happy to re-investigate this. If the in-between state does still happen, your approach in this PR sounds like a valid fix to me.

jhuleatt:
Closing this one out since I think it has been addressed. Please open a new issue if you'd like to discuss reviving it.

davidhewitt:
Thanks for the update; I have no longer got access to the original codebase where I was experiencing issues. From your comments above I'm similarly convinced that the original issue here is solved.

PR #252, opened 2020-05-03
Add useFirestoreCollectionDataOnce

banyan:
A part of https://github.com/FirebaseExtended/reactfire/issues/137

I've implemented this with reference to https://github.com/FirebaseExtended/reactfire/pull/211 :)

### DIFF
--- a/reactfire/firestore/index.tsx
+++ b/reactfire/firestore/index.tsx
@@ -114,3 +114,16 @@ export function useFirestoreCollectionData<T = { [key: string]: unknown }>(query
 
   return useObservable(collectionData(query, idField), queryId, checkStartWithValue(options));
 }
+
+/**
+ * Get a firestore collection and don't subscribe to changes
+ *
+ * @param ref - Reference to the collection you want to get
+ * @param options
+ */
+export function useFirestoreCollectionDataOnce<T = { [key: string]: unknown }>(query: firestore.Query, options?: ReactFireOptions<T[]>): T[] {
+  const idField = checkIdField(options);
+  const queryId = `firestore:collectionDataOnce:${getUniqueIdForFirestoreQuery(query)}:idField=${idField}`;
+
+  return useObservable(collectionData(query, idField).pipe(first()), queryId, checkStartWithValue(options));
+}
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -27,6 +27,7 @@
       - [`useFirestoreDocDataOnce`](#useFirestoreDocDataOnce)
       - [`useFirestoreCollection`](#useFirestoreCollection)
       - [`useFirestoreCollectionData`](#useFirestoreCollectionData)
+      - [`useFirestoreCollectionDataOnce`](#useFirestoreCollectionDataOnce)
     - Realtime Database
       - [`useDatabaseObject`](#useDatabaseObject)
       - [`useDatabaseObjectData`](#useDatabaseObjectData)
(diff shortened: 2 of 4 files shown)


banyan:
@googlebot I signed it!

reactfire/firestore/index.tsx
```diff
@@ -114,3 +114,16 @@ export function useFirestoreCollectionData<T = { [key: string]: unknown }>(query
...
+ * Get a firestore collection and don't subscribe to changes
+ *
+ * @param ref - Reference to the collection you want to get
+ * @param options
+ */
+export function useFirestoreCollectionDataOnce<T = { [key: string]: unknown }>(query: firestore.Query, options?: ReactFireOptions<T[]>): T[] {
+  const idField = checkIdField(options);
+  const queryId = `firestore:collectionDataOnce:${getUniqueIdForFirestoreQuery(query)}:idField=${idField}`;
+
+  return useObservable(collectionData(query, idField).pipe(first()), queryId, checkStartWithValue(options));
```
davideast, line 128: This is tricky. I'm not sure if using `first()` to terminate an `onSnapshot()` listener is the right call when we could use `.get()` which is meant for one time reads.

What's tricky is that we don't support this natively in RxFire and ReactFire does not take a dependency of RxJS. It's actually not that hard to do with RxFire. 

```ts
function getCollectionData(colRef) {
  return from(citiesCol.get());
}
```
 So maybe we just add that to RxFire and then add that here? What are you thoughts?

davideast (commented):
Hey @banyan! 

Thank you so much for the PR with the docs and the tests! I have one small question about the implementation, so let me know your thoughts.

reactfire/firestore/index.tsx
banyan: Thanks for the insight! 

>we could use .get() which is meant for one time reads.

I totally agree. 

>ReactFire does not take a dependency of RxJS.

Maybe I don't understand it properly, but it looks like it does have [the dependency on rxjs](https://github.com/FirebaseExtended/reactfire/blob/b82b58a8146eb044321244005f9d0eeeaf2be9e1/reactfire/package.json#L29).
So it looks like I could write the following if I want, but maybe it's just not a good thing design-wise for the reactfire and rxfire libraries (I'm not sure).

```diff
diff --git a/reactfire/firestore/index.tsx b/reactfire/firestore/index.tsx
index 0974b14..14971b1 100644
--- a/reactfire/firestore/index.tsx
+++ b/reactfire/firestore/index.tsx
@@ -1,9 +1,10 @@
 import { firestore } from 'firebase/app';
-import { collectionData, doc, docData, fromCollectionRef } from 'rxfire/firestore';
+import { collectionData, doc, docData, fromCollectionRef, snapToData } from 'rxfire/firestore';
 import { preloadFirestore, ReactFireOptions, useObservable, checkIdField, checkStartWithValue } from '..';
 import { preloadObservable } from '../useObservable';
-import { first } from 'rxjs/operators';
+import { map, first } from 'rxjs/operators';
 import { useFirebaseApp } from '../firebaseApp';
+import { from } from 'rxjs';
 
 const CACHED_QUERIES = '_reactFireFirestoreQueryCache';
 
@@ -124,6 +125,6 @@ export function useFirestoreCollectionData<T = { [key: string]: unknown }>(query
 export function useFirestoreCollectionDataOnce<T = { [key: string]: unknown }>(query: firestore.Query, options?: ReactFireOptions<T[]>): T[] {
   const idField = checkIdField(options);
   const queryId = `firestore:collectionDataOnce:${getUniqueIdForFirestoreQuery(query)}:idField=${idField}`;
-
-  return useObservable(collectionData(query, idField).pipe(first()), queryId, checkStartWithValue(options));
+  const onetimeCollectionData = from(query.get()).pipe(map(snapshot => snapshot.docs.map(snap => snapToData(snap, idField) as T)));
+  return useObservable(onetimeCollectionData, queryId, checkStartWithValue(options));
 }
```

If you'd like me to put some logic on the rxfire side, I'd be happy to do so! Please let me know. Thanks!

steurt:
Would be great to have this `useFirestoreCollectionDataOnce`. Any progress @banyan ?

reactfire/firestore/index.tsx
steurt: @davideast and @banyan any update on this?
It would be a great addition to reactfire.

n-sviridenko:
Hey guys. Any progress here? This seems to be pretty handy to have.

tyler-reitz:
Thanks for this contribution! The feature itself is valid and something we'd like to support. However, this PR was built against the v8 Firebase API and is no longer compatible with the current codebase. Rather than ask you to do a full rewrite, we're going to close this and track the feature in #473. We hope to address it as part of the v5 roadmap. Thanks again for taking the time!

PR #253, opened 2020-05-04
Update preload methods in sample

davideast:
Update sample app in the repository to use the latest API for `preloadFactory()` methods:

```ts
preloadFirestore({
   firebaseApp,
   setup(firestore) {
     return firestore().enablePersistence();
   }
}),
```

### DIFF
--- a/sample/src/App.js
+++ b/sample/src/App.js
@@ -38,20 +38,29 @@ const Card = ({ title, children }) => {
 // fetching them now
 const preloadSDKs = firebaseApp => {
   return Promise.all([
-    preloadFirestore(firebaseApp, firestore => {
-      return firestore().enablePersistence();
+    preloadFirestore({
+      firebaseApp,
+      setup(firestore) {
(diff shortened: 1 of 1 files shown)


sample/src/App.js
```diff
@@ -38,20 +38,29 @@ const Card = ({ title, children }) => {
 // fetching them now
 const preloadSDKs = firebaseApp => {
   return Promise.all([
-    preloadFirestore(firebaseApp, firestore => {
-      return firestore().enablePersistence();
+    preloadFirestore({
+      firebaseApp,
+      setup(firestore) {
```
jhuleatt, line 43: any opinion on 

```js
setup(firestore) {
```

vs

```js
setup: firestore => {
```

for all of these?

The second is a bit more clear to me, but that could just be me

davideast: I prefer the named function. AFIK, it will show up in the stack trace and the arrow function will not.

PR #254, opened 2020-05-04
Add version number as an export

davideast:
## Version number in export

```ts
import { version } from 'reactfire';
```

Also added version registration in core sdk.

### DIFF
--- a/reactfire/after.build.js
+++ b/reactfire/after.build.js
@@ -5,7 +5,7 @@ const { version } = require('./package.json');
 // This script should only ever be ran after
 // a successful rollup build.
 
-const ABSOULTE_PATH = path.join(process.cwd(), `/pub/reactfire`);
+const ABSOLUTE_PATH = path.join(process.cwd(), `/pub/reactfire`);
 
 const files = [{
   name: 'firebaseApp/index.js',
--- a/reactfire/after.build.js
+++ b/reactfire/after.build.js
@@ -14,7 +14,7 @@ const files = [{
 }];
 
 const fileContents = files.map(file => {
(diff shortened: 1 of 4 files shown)


reactfire/firebaseApp/index.tsx
```diff
@@ -18,11 +16,13 @@ type Props = {
   appName?: string;
 };
 
+const version = '::__reactfireversion__::';
```
jhuleatt, line 19:
```suggestion

// The version number is substituted in as part of the build process
// See after.build.js for the substitution script
const version = '::__reactfireversion__::';
```

reactfire/after.build.js
```diff
@@ -0,0 +1,29 @@
+const fs = require('fs');
+const path = require('path');
+const { version } = require('./package.json');
+
+// This script should only ever be ran after
+// a successful rollup build.
+
+const ABSOULTE_PATH = path.join(process.cwd(), `/pub/reactfire`);
```
jhuleatt, line 8:
```suggestion
const ABSOLUTE_PATH = path.join(process.cwd(), `/pub/reactfire`);
```

reactfire/after.build.js
```diff
@@ -0,0 +1,29 @@
...
+
+const files = [{
+  name: 'firebaseApp/index.js',
+  token: '::__reactfireversion__::',
+  replaceValue: version
+}];
+
+const fileContents = files.map(file => {
+  const fullPath = path.join(ABSOULTE_PATH, file.name);
```
jhuleatt, line 17:
```suggestion
  const fullPath = path.join(ABSOLUTE_PATH, file.name);
```

PR #255, opened 2020-05-13
Optional suspense

jhuleatt:
### What's happening?

This is a proposed API change that makes it easier to use ReactFire without enabling experimental React features (Suspense & Concurrent Mode).

"Suspense Mode" will be settable in `FirebaseAppProvider` (in which case it will automatically apply to all children of `FirebaseAppProvider`), or in individual calls as part of `ReactFireOptions`.

```jsx
<FirebaseAppProvider firebaseConfig={config} suspense={true}>
    <App />
</FirebaseAppProvider>
```

**If you don't set `suspense={true}` in `FirebaseAppProvider`, it will default to `false`.**

Now that Suspense won't necessarily be handling loading states, each hook will return more information about its status:

```ts
export interface ObservableStatus<T> {
  status:
    | 'loading' // waiting for first value from observable
    | 'error'
    | 'success'; // has received at least one value
  hasEmitted: boolean; // has received at least one value
  isComplete: boolean; // observable has triggered onComplete event
  data: T; // latest data from observable
  error: Error | undefined;
  firstValuePromise: Promise<void>; // promise that resolves after first emit from observable
}
```

### Status

- [x] Make API changes to core library
- [x] Update tests for new API, and add new tests that don't use Suspense
- [x] Make a sample that works without Suspense
- [x] Ensure all Suspense samples and tests still work
- [x] Update docs with new API:
    - [x] README
    - [x] Guide (just added a disclaimer that samples are for Concurrent Mode)
    - [x] Quickstart (just added a disclaimer that samples are for Concurrent Mode)
    - [x] Reference


zaccharles:
Thanks for the hard work, is there anything I can do to help get this across the line?

praneybehl:
Seems like one of the pipelines is stuck since a few months. Is this project dead?

kantmichel:
i'd be very keen on using this in the current stable react version. Any idea's on when this is being merged?

mastef:
@jhuleatt @jamesdaniels Any chance this proposal may make it any time soon into ReactFire? Since it was mentioned in #260 and there's great interest in using ReactFire with the stable builds of React.

Thank you for your work on this!

jhuleatt:
Hey there, apologies for the long silence on this. We're working on reviving this PR, and I've just pushed the latest updates. 

Examples of the new api can be found here: https://github.com/FirebaseExtended/reactfire/tree/optional-suspense/example/withoutSuspense

If you want to pull this branch yourself and try things out, pull it then run:

In one terminal:
```shell
yarn
yarn start
```

In a second terminal:
```shell
cd example
yarn
yarn start
```

src/database.tsx
```diff
@@ -1,16 +1,19 @@
...
-import { ReactFireOptions, useObservable, checkIdField, checkStartWithValue } from '..';
+import { ReactFireOptions, useObservable, checkIdField, ObservableStatus } from './';
 
 import { Observable } from 'rxjs';
 import { map } from 'rxjs/operators';
 
 const CACHED_QUERIES = '_reactFireDatabaseCachedQueries';
 
 // Since we're side-effect free, we need to ensure our observableId cache is global
+// @ts-ignore: TS doesn't like globalThis
```
jamesdaniels, line 11: globalThis should have come with typescript 3.4 https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html, is there somethin missing in the tsconfig?

jhuleatt: I get the error:

```
Property '_reactFireDatabaseCachedQueries' does not exist on type 'typeof globalThis'
```

Maybe I need to find a way to declare the keys we expect to be on `globalThis`?

startupgurukul:
Hello folks, 
Thank you for this, Looking forward to being able to use it in production builds of react.

src/database.tsx
jhuleatt: Resolving since we figured this out together. We now have a `ReactFireGlobals` type and we just cast `globalThis`:

```ts
((globalThis as any) as ReactFireGlobals)
```

PR #274, opened 2020-08-10
fix: issue with _warmupSubscription being undefined

zhigang1992:
### Description

Bug fixes, this is not just happening in test but also in react-native env.

### Code sample

I've no idea how this fix would break that current test.

I've verified that the test is indeed failing because of this change.

But I have no idea why or how to fix / correct this test.

Please help.

### DIFF
--- a/reactfire/useObservable/SuspenseSubject.ts
+++ b/reactfire/useObservable/SuspenseSubject.ts
@@ -35,7 +35,7 @@ export class SuspenseSubject<T> extends Subject<T> {
 
     // set a timeout for reseting the cache, subscriptions will cancel the timeout
     // and reschedule again on unsubscribe
-    this._timeoutHandler = setTimeout(this._reset, this._timeoutWindow);
+    this._timeoutHandler = setTimeout(this._reset.bind(this), this._timeoutWindow);
   }
 
   get hasValue(): boolean {
@@ -66,10 +66,7 @@ export class SuspenseSubject<T> extends Subject<T> {
   }
 
   private _reset() {
-    // seems to be undefined in tests?
-    if (this._warmupSubscription) {
-      this._warmupSubscription.unsubscribe();
-    }
+    this._warmupSubscription.unsubscribe();
     this._hasValue = false;
     this._value = undefined;
     this._error = undefined;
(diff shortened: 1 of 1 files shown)


zhigang1992:
@googlebot I signed it!

zhigang1992:
I've no idea how this fix would break that current test.

I've verified that the test is indeed failing because of this change.

But I have no idea why or how to fix / correct this test.

Please help.

jamesdaniels (approved):
LGTM, thanks

jamesdaniels:
Closing in favor of #300

jamesdaniels (changes requested):
Looks like these are failing with the new tests / changes on master. Can you take a peek?

zhigang1992:
Yeah, as I've posted above in the description. I need help with that. I'm not sure how to fix it

fhervieux:
Hi! I had a look at this since I stumbled upon the issue in a react native app. I think the tests are failing because now that the reset method is bound the subject value is erased as soon as it is emitted.

We add a call to reset to the inner observable subscription here: https://github.com/FirebaseExtended/reactfire/blob/main/src/SuspenseSubject.ts#L87
So the subject never has a value when we look for it here: https://github.com/FirebaseExtended/reactfire/blob/main/src/sdk.tsx#L52

That is what is making the test fail. The first emission will reset the subject so a promise is thrown even when it was actually resolved once and the suspense never renders its component.

Removing this (changing `return this._innerSubscriber.add(this._reset.bind(reset));` to `return this._innerSubscriber` in SuspenseSubject#_subscribe) and keeping the bound call in the timeout handler fixes the tests and the app still behaves as expected but I don't understand why this call was there in the first place so I might be missing something completely! Can someone confirm that this would not introduce a regression?

Kurtoid:
Is there any short-term workaround for react-native on expo?

Kurtoid:
I was thinking about it, but I didn't know if it would cause any side effects with Suspense (which Im not using for my project)

I can publish it, but I can't guarantee it'll work reliably. I've only used/tested a small set of the reactfire features

Kurtoid:
It's on my profile as kurtoid/reactfire

You should be able to pull it in through NPM as `@kurtoid/reactfire`

olawalejuwonm:
okay i'll check it out

jhuleatt:
@zhigang1992 thanks for this, I've included this fix in PR #372 because I was making a bunch of other changes to `SuspenseSubject`. Since you originally proposed this fix, I gave you co-author credit in that PR.

PR #306, opened 2020-12-07
Some updates related to Optional Suspense

steveasher:
### Description

I'm setting up a new project, and I noticed that the docs for `useUser()` are not up-to-date with the latest from #255.  I haven't spent time looking for similar discrepancies for other hooks.  I just figured I should fix this one while it is right in front of me.


jhuleatt (approved):
Thanks so much for the docs help, @steveasher!

PR #307, opened 2020-12-08
Reference docs improvements

jhuleatt:
- Change script to generate links to master instead of a specific git commit
    - Regenerate reference docs so they all point to master
- Add a new GH Workflow that checks if reference docs need to be regenerated (I want to add this as a required check for every PR)

Review notes: This is a noisy PR. Ignore all `.md` files, just take a look at `package.json` and `docs.yaml`


jamesdaniels (approved):
LGTM

PR #310, opened 2020-12-11
Mark startWithValue deprecated and alias it to initialData

### LINKED ISSUE #297: Alias startWithValue to initialData
#255 changed the `startWithValue` option to `initialData`. Let's alias `startWithValue` and mark it deprecated in our types so that it's easier to migrate.

jhuleatt:
### Description

Fixes #297 

We switched from `startWithValue` to `initialData` in ReactFire v3. This PR marks `startWithValue` deprecated instead of fully removing it, and checks for `startWithValue` and uses that if `initialData` isn't available.

Another option is to *only* mark `startWithValue` deprecated, and not check for it anywhere.

### DIFF
--- a/src/storage.tsx
+++ b/src/storage.tsx
@@ -41,7 +41,7 @@ export function useStorageTask<T = unknown>(
   const observableId = `storage:task:${ref.toString()}`;
   const observable$ = _fromTask(task);
 
-  return useObservable(observableId, observable$, options ? options.initialData : undefined);
+  return useObservable(observableId, observable$, options ? options.initialData ?? options.startWithValue : undefined);
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/storage.tsx
```diff
@@ -41,7 +41,7 @@ export function useStorageTask<T = unknown>(
   const observableId = `storage:task:${ref.toString()}`;
   const observable$ = _fromTask(task);
 
-  return useObservable(observableId, observable$, options ? options.initialData : undefined);
+  return useObservable(observableId, observable$, options ? options.initialData ?? options.startWithValue : undefined);
```
jamesdaniels, line 44: Before and after the change this doesn't look correct to me, shouldn't it just pass options? `useObservable`'s third argument is an object.

jamesdaniels (approved):
LGTM, note feedback on storage but it appears that was wrong before this PR.

src/storage.tsx
jhuleatt: great catch, fixed

PR #318, opened 2020-12-30
feat: added loader fallback to AuthCheck

prilutskiy:
### Description

When `<AuthCheck />` does its check and is in a loading state, we see a fallback component on the screen for a couple of moments. It would be nice to have an opportunity to see a loader instead.

### Code sample

```jsx
<AuthCheck
  fallback={<AuthView />}
  loader={<div>App is loading...</div>}
>
  <MyApp />
</AuthCheck>
```


rayshan:
Would love to have this. Without it the Flash of Unauthenticated UI is a bad experience for the user.

jhuleatt:
Hey @prilutskiy, thanks for the PR.  I've added some context around why we're slow on responding to this in https://github.com/FirebaseExtended/reactfire/issues/325#issuecomment-827654376

PR #326, opened 2021-01-14
Change quickstart to support stable builds of React

### LINKED ISSUE #321: Update quickstart to 3.x api
The [quickstart](https://github.com/FirebaseExtended/reactfire/blob/master/docs/quickstart.md) doesn't use the latest 3.x apis. We need to update it to support them. At the same time, we should consider making sure it works without concurrent mode.

jhuleatt:
Fixes #321 

### Description

Remove concurrent mode features from quickstart so that it works with stable builds of React

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Depending on your targeted platforms you may need to install polyfills. The most
 ## Docs
 
 - [**Quickstart**](./docs/quickstart.md)
+  - Advanced: If you're using Concurrent Mode, check out the [Concurrent Mode quickstart](./docs/quickstart-concurrent-mode.md)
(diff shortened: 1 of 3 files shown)


ngocdaothanh:
@jhuleatt I'm new to reactfire. I've read the code in your PR. Can you please explain why in `docs/quickstart.md` you write this:

```
const { status, data } = useFirestoreDocData(burritoRef);
```

But in `docs/quickstart-concurrent-mode.md` you write this:

```
const burrito = useFirestoreDocData(burritoRef);
```

Why the same function `useFirestoreDocData` returns 2 different results?

Issue #325, opened 2021-01-11
ClaimsCheck throwing TypeError: Cannot read property 'claims' of undefined


thomasmburke:
### Version info

**React:** ^16.14.0

**Firebase:** 8.2.1

**ReactFire:** 3.0.0-rc.0

### Steps to reproduce

Seems to happen with any `AuthCheck` that uses the `requiredClaims` param or by calling `ClaimsCheck` directly
```jsx
<AuthCheck fallback={<Home />} requiredClaims={{admin: true}}>
    <Admin />
</AuthCheck>}
```

### Expected behavior

Users with custom claims should pass the `AuthCheck` or `ClaimsCheck` and fallback to the fallback argument

### Actual behavior

Receiving error: `TypeError: Cannot read property 'claims' of undefined`

![image](https://user-images.githubusercontent.com/40719837/104142674-68d2fe00-5371-11eb-92d4-ae3e0ed7f1c5.png)

Debugging in Chrome Dev tools indicates that the `useObservable` call may have an unresolved promise? I'm guessing this because `_useIdTokenResult` has a status of "loading" and `isComplete` is false.

![image](https://user-images.githubusercontent.com/40719837/104142807-f57dbc00-5371-11eb-8f0a-0f95a2a5c7ee.png)

Additionally, when I expand the user variable that I have blacked out I can see that it successfully is getting my authenticated user, meaning I can see the email, displayName, and custom claims when inspecting the JWT token. The same `<AuthCheck>` call works if I omit the `requiredClaims` param.

Ljugoboss:
I'm running into the same issue.

The problem seems to occur due to `latest` always ending up defaulting to `config.initialData` in `useObservable` the first time it returns a value, but no config is sent in from `CheckClaims`;
![image](https://user-images.githubusercontent.com/4201301/105459552-b7926900-5c8a-11eb-9a94-3ae3180acc29.png)
![image](https://user-images.githubusercontent.com/4201301/105459171-0f7ca000-5c8a-11eb-91a4-fda8c5e15e66.png)

I solved it locally by adding default values to `data` and `claims`, or by actually sending in a config with `initialData`.
![image](https://user-images.githubusercontent.com/4201301/105459435-89148e00-5c8a-11eb-970a-59b1333515d8.png)
![image](https://user-images.githubusercontent.com/4201301/105459820-2a9bdf80-5c8b-11eb-9734-68434dd68726.png)

ElinOlundForsling:
[...]

PR #336, opened 2021-02-12
Provide an initial value for the claims object (#325)

mjakl:
This PR captures the proposed fix for #325, i.e. providing a initialisation value for `claims`.

This prevents `TypeError: Cannot read property 'claims' of undefined` errors. This happened during tests using jsdom in my case, it was working fine in the Browser (Chrome).

### DIFF
--- a/src/auth.tsx
+++ b/src/auth.tsx
@@ -72,7 +72,7 @@ export interface ClaimsCheckProps {
 }
 
 export function ClaimsCheck({ user, fallback, children, requiredClaims }: ClaimsCheckProps) {
-  const { data } = useIdTokenResult(user, false);
+  const { data } = useIdTokenResult(user, false, { initialData: { claims: {} } });
   const { claims } = data;
   const missingClaims: { [key: string]: { expected: string; actual: string } } = {};
 


mjakl:
@googlebot I signed it!

webbertakken:
I just ran into the same issue. I reproduced the fix in this PR and can confirm that it works!

Looking forward to see this merged.

mjakl:
Note that by providing `initialData` the first version of the PR broke the suspense functionality. I've updated the PR to deal with the undefined claims object instead of providing an `initialData` object.

aalises:
Ran into the same issue! would be great to have it merged.

jhuleatt:
Hey @mjakl, thanks for the PR.  I've added some context around why we're slow on responding to this in https://github.com/FirebaseExtended/reactfire/issues/325#issuecomment-827654376

webbertakken:
Would merging this as a temporary measure be useful to unblock implementers?

jhuleatt:
I'd prefer not to merge this, and instead move toward the changes in #368 (deprecating `AuthCheck` and `ClaimsCheck` in favor of a hook that covers their functionality). If you disagree, please add your thoughts to #368

PR #367, opened 2021-04-30
Obey the Rules of Hooks

### LINKED ISSUE #337: Suggested signature cleanup in example; side effect vs. hook
Hi there, and thanks for this great project! 
I especially like how it keeps the signature for small projects very tidy.

On topic; I was looking through the code and some of the examples and spotted this [todo](https://github.com/FirebaseExtended/reactfire/blob/252b60781086d657e4d43fa12237e0a068e395f1/src/auth.tsx#L9) and another [todo](https://github.com/FirebaseExtended/reactfire/blob/252b60781086d657e4d43fa12237e0a068e395f1/src/auth.tsx#L25). To practice I had a look at it and thought about how I would solve this if it were my project. I'm here to learn so I'd love to hear your feedback on the approach.

#### Problem

The problem here seems to be about some functions that have the signature of both a hook and as a regular function outside of the react tree, injecting context through a parameters. I would argue this on itself is an anti-pattern, and the solution may be simple if I'm not mistaken;

#### Case `preloadUser`

Looking at `preloadUser` we're dealing with a side effect, that would always be called from `useEffect` and never as a contextual hook like `usePreloadUser`. As far as I can see it's not used as a hook now either.

Suggested change to [preloadUser](https://github.com/FirebaseExtended/reactfire/blob/252b60781086d657e4d43fa12237e0a068e395f1/src/auth.tsx#L8)

```diff
- export function preloadUser(options?: { firebaseApp?: firebase.app.App }) {
-    // TODO: Find an alternative that doesn't break the rules of hooks (conditional hook call)
-   const firebaseApp = options?.firebaseApp || useFirebaseApp();
- 
+ export function preloadUser(firebaseApp: firebase.app.App) {
    return preloadAuth({ firebaseApp }).then(auth => {
      const result = preloadObservable(user(auth()), `auth:user:${firebaseApp.name}`);
      return result.toPromise();
    });
  }
```

#### Case `useUser`

- For `useUser` the opposite applies, it would always be used as a hook en never need to be provided an auth object.

There is one reference that violates this signature;

image

Since the reference is a component that, I would argue that it would realistically never used outside of `FirebaseAppProvider`, and suggest the following changes:

to [AuthCheck](https://github.com/FirebaseExtended/reactfire/blob/252b60781086d657e4d43fa12237e0a068e395f1/src/auth.tsx#L97)
```diff
- export function AuthCheck({ auth, fallback, children, requiredClaims }: AuthCheckProps): JSX.Element {
+ export function AuthCheck({ fallback, children, requiredClaims }: AuthCheckProps): JSX.Element {
  const { data: user } = useUser();

  if (user) {
    return requiredClaims ? (
      <ClaimsCheck user={user} fallback={fallback} requiredClaims={requiredClaims}>
        {children}
      </ClaimsCheck>
    ) : (
      <>{children}</>
    );
  } else {
    return <>{fallback}</>;
  }
}
```

to [useUser](https://github.com/FirebaseExtended/reactfire/blob/252b60781086d657e4d43fa12237e0a068e395f1/src/auth.tsx#L24)
```diff
/**
 * Subscribe to Firebase auth state changes, including token refresh
 *
- * @param auth - the [firebase.auth](https://firebase.google.com/docs/reference/js/firebase.auth) object
 * @param options
 */
- export function useUser<T = unknown>(auth?: firebase.auth.Auth, options?: ReactFireOptions<T>): ObservableStatus<firebase.User> {
-  // TODO: Find an alternative that doesn't break the rules of hooks (conditional hook call)
-  auth = auth || useAuth();
-
-  if (!auth) {
-    throw new Error('firebase.auth not found');
-  }
+ export function useUser<T = unknown>(options?: ReactFireOptions<T>): ObservableStatus<firebase.User> {
+  auth = useAuth();

  const observableId = `auth:user:${auth.app.name}`;
```

Finally the error could be shown in `useAuth();` as you would always expect `firebase.auth` to be in scope.

If you agree with these changes I'd be happy to send a PR. If not, I'd love to learn what mistakes I'm making.
Thank you for your consideration.

jhuleatt:
### Description

fixes #337 

This is a breaking change to make ReactFire comply with the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html). 

- `firebaseApp` is no longer optional in the `preload*` functions - you have to pass it in
- You can no longer pass in your own `firebaseApp` instance to hooks
- You can no longer pass in your own `auth` instance  to `AuthCheck`

### DIFF
--- a/docs/reference/interfaces/auth.authcheckprops.md
+++ b/docs/reference/interfaces/auth.authcheckprops.md
@@ -8,26 +8,17 @@
 
 ### Properties
 
-- [auth](auth.authcheckprops.md#auth)
 - [children](auth.authcheckprops.md#children)
 - [fallback](auth.authcheckprops.md#fallback)
 - [requiredClaims](auth.authcheckprops.md#requiredclaims)
 
 ## Properties
 
-### auth
-
-• `Optional` **auth**: Auth
-
-Defined in: [src/auth.tsx:61](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L61)
-
-___
-
 ### children
 
 • **children**: ReactNode
 
-Defined in: [src/auth.tsx:63](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L63)
+Defined in: [src/auth.tsx:54](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L54)
 
 ___
 
@@ -35,7 +26,7 @@ ___
 
 • **fallback**: ReactNode
 
-Defined in: [src/auth.tsx:62](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L62)
+Defined in: [src/auth.tsx:53](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L53)
 
 ___
 
@@ -43,4 +34,4 @@ ___
 
 • `Optional` **requiredClaims**: Object
 
-Defined in: [src/auth.tsx:64](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L64)
+Defined in: [src/auth.tsx:55](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L55)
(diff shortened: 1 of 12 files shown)


jamesdaniels (approved):
LGTM

PR #368, opened 2021-04-30
New Auth Hook: `useSigninCheck()` to replace `AuthCheck` and `ClaimsCheck` components

### LINKED ISSUE #325: ClaimsCheck throwing TypeError: Cannot read property 'claims' of undefined

jhuleatt:
Fixes #325 

### Description

Deprecate `AuthCheck` and `ClaimsCheck` in favor of a hook that would allow easy implementation of either of those components: **`useSigninCheck()`**

### Code sample

Simple use case:

```jsx
function UserFavorites() {
   const {status, data: signInCheckResult} = useSigninCheck();

   if (status === 'loading') {
     return <LoadingSpinner />
   }

   if (signInCheckResult.signedIn === true) {
     return <FavoritesList />
   } else {
     return <SignInForm />
   }
}
```

Advanced: You can also optionally check [custom claims](https://firebase.google.com/docs/auth/admin/custom-claims). Example:

```jsx
function ProductPricesAdminPanel() {
  const { status, data: signInCheckResult } = useSigninCheck({
    requiredClaims: { admin: true, canModifyPrices: true },
  });

  if (status === "loading") {
    return <LoadingSpinner />;
  }

  if (signInCheckResult.signedIn && signInCheckResult.hasRequiredClaims) {
    return <FavoritesList />;
  } else {
    console.warn(
      `User ${signInCheckResult.user.displayname} is missing claims:`,
      Object.keys(signInCheckResult.errors)
    );
    return <SignInForm />;
  }
}
```

### Why deprecate `AuthCheck` and `ClaimsCheck`?

They worked well in the previous version of ReactFire when ReactFire only supported concurrent mode, but are awkward now that we support both concurrent mode and non-concurrent mode (added in https://github.com/FirebaseExtended/reactfire/pull/255). Moving to a hook allows for the same easy checks, but leaves loading states to the developer, instead of `AuthCheck` and `ClaimsCheck` trying to handle both concurrent mode and non-concurrent mode loading strategies.

### DIFF
--- a/src/auth.tsx
+++ b/src/auth.tsx
@@ -64,7 +64,7 @@ export interface ClaimsCheckProps {
 }
 
 interface ClaimCheckErrors {
-  [key: string]: ReactFireError[];
+  [key: string]: any[];
 }
 
 export type SigninCheckResult =
--- a/src/auth.tsx
+++ b/src/auth.tsx
@@ -63,8 +63,8 @@ export interface ClaimsCheckProps {
   requiredClaims: { [key: string]: any };
 }
 
-interface MissingClaims {
-  [key: string]: { expected: string; actual: string };
+interface ClaimCheckErrors {
+  [key: string]: String[];
 }
 
 export type SigninCheckResult =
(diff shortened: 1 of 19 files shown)


src/auth.tsx
```diff
@@ -68,9 +69,103 @@ export interface ClaimsCheckProps {
...
+                  actual: idTokenResult.claims[claim]
+                };
+              }
+            });
+
+            return { signedIn: true, hasRequiredClaims: Object.keys(missingClaims).length === 0, missingClaims };
+          })
+        );
+      } else {
+        return of({ signedIn: true });
```
jamesdaniels, line 149: Should we return `hasRequiredClaims: true` here too, just so it's more strongly typed, like the `!user` case. Though perhaps we shouldn't return `hasRequiredClaims` if they don't provide claims, in that case we should overload the definition of the function, to more strongly enforce wether `hasRequiredClaims` should be expected on the return object.

jamesdaniels (approved):
LGTM though look into the comment I left.

src/auth.tsx
jhuleatt: I went back and forth on the same thing. I guess if you require zero claims, then `hasRequiredClaims` is technically true. I'll add it in

jhuleatt: Added

jhuleatt:
Open question: Should the `SigninCheckResult` object return the current user too? Currently it has:

```ts
interface MissingClaims {
  [key: string]: { expected: string; actual: string };
}

export type SigninCheckResult =
  | {
      signedIn: false;
      hasRequiredClaims: false;
    }
  | { signedIn: true; hasRequiredClaims: boolean; missingClaims?: MissingClaims };
```

jhuleatt:
> Open question: Should the `SigninCheckResult` object returned by `useSigninCheck` contain the current user too? 
> 
> Or should we leave that to `useUser`?

After a chat with @thomasmburke, I think the answer is that we should add it. `useSigninCheck` gets the current user anyway so that it can check claims, so it will be trivial to implement. And there may be some use case where it's nice to not have to call `useUser` alongside `useSigninCheck`.

src/auth.tsx
```diff
@@ -59,9 +60,107 @@ export interface ClaimsCheckProps {
...
+                };
+              }
+            });
+
+            return { signedIn: true, hasRequiredClaims: Object.keys(missingClaims).length === 0, missingClaims };
+          })
+        );
+      } else {
+        return of({ signedIn: true, hasRequiredClaims: true });
+      }
```
jhuleatt, line 150: @jamesdaniels or @davideast am I rxjs-ing correctly here? Any suggested improvements?

test/auth.test.tsx
```diff
@@ -123,6 +127,67 @@ describe('Authentication', () => {
...
+      });
+
+      // Signed in
+      expect(app.auth().currentUser).not.toBeNull();
+      expect(result.current.data).toEqual({ signedIn: true, hasRequiredClaims: true, user: app.auth().currentUser });
+    });
+
+    // Skipping this because of an issue setting custom claims
+    // https://github.com/firebase/firebase-tools/issues/3083
+    it.skip('checks custom claims', async () => {
```
jhuleatt, line 151: skipping this test because of https://github.com/firebase/firebase-tools/issues/3083

aalises:
Amazing! How would you go about allowing a claim to be a set of values? I have this `role` claim that can be a fixed set of values e.g. `ADMIN, CUSTOMER`.

I am using a custom implementation of `ClaimsCheck` where you can allow an array of values to be passed for a claim and validates if one of them passes. 

For example, I would want a feature to only be available for admins and also some other special users, so I could do:

```
{role: ["ADMIN", "SPECIAL_USER"]}
```

whereas in the current implementation only one value can be checked by claim... how would you guys contemplate this case?

jhuleatt:
> How would you go about allowing a claim to be a set of values? I have this `role` claim that can be a fixed set of string values e.g. `ADMIN, CUSTOMER`.
>
> In the current implementation only one value can be checked by custom claim... how would you guys contemplate this case?

Thanks for the thoughtful feedback @aalises! Maybe we should allow a validator function to be passed to `useSigninCheck` as an alternative to a `requiredClaims` object if more advanced validation logic is needed. Something like:

```ts
const { status, data: signInCheckResult } = useSigninCheck({
  checkClaims: (userClaims) => {
    const validClaimsSet = ["ADMIN", "SPECIAL_USER"];
    let hasAnyClaim = false;

    for (const claim of validClaimsSet) {
      if (userClaims[claim] === true) {
        hasAnyClaim = true;
        break;
      }
    }

    // To populate the `SigninCheckResult` object
    return {
      hasRequiredClaims: hasAnyClaim,
      missingClaims: hasAnyClaim ? undefined : validClaimsSet,
    };
  },
});
```

What do you think?

aalises:
That'd be great @jhuleatt !

src/auth.tsx
jhuleatt: I'm more confident in this now that tests are passing

test/auth.test.tsx
jhuleatt: fixed this

src/auth.tsx
```diff
@@ -59,9 +60,136 @@ export interface ClaimsCheckProps {
...
+  const auth = useAuth();
+
+  // ObservableId should change for different options configurations to ensure no cache collisions
+  let observableId = `auth:signInCheck:${auth.app.name}::forceRefresh:${!!options?.forceRefresh}`;
+  if (options?.forceRefresh) {
+    observableId = `${observableId}:forceRefresh:${options.forceRefresh}`;
+  }
+  if (options?.hasOwnProperty('requiredClaims')) {
+    observableId = `${observableId}:requiredClaims:${JSON.stringify((options as SignInCheckOptionsClaimsObject).requiredClaims)}`;
+  } else if (options?.hasOwnProperty('validateCustomClaims')) {
```
jhuleatt, line 136: note: IE11 might not support stringifying a function

src/auth.tsx
```diff
@@ -59,9 +60,136 @@ export interface ClaimsCheckProps {
...
 }
 
+interface ClaimCheckErrors {
+  [key: string]: String[];
+}
+
+export type SigninCheckResult =
+  | {
+      signedIn: false;
+      hasRequiredClaims: false;
```
jhuleatt, line 73: consider adding empty errors here to make destructuring easier

jhuleatt: or null?

src/auth.tsx
```diff
@@ -59,9 +60,136 @@ export interface ClaimsCheckProps {
...
+  return useObservable(observableId, observable);
+}
+
+function getClaimsObjectValidator(requiredClaims: firebase.auth.IdTokenResult['claims']): ClaimsValidator {
+  return function claimsObjectValidator(userClaims) {
+    const errors: { [key: string]: String[] } = {};
+
+    Object.keys(requiredClaims).forEach(claim => {
+      if (requiredClaims[claim] !== userClaims[claim]) {
+        errors[claim] = [`Expected "${requiredClaims[claim]}", but user has "${userClaims[claim]}" instead`];
```
jhuleatt, line 175: change to `new ClaimsCheckError`

jhuleatt: https://github.com/firebase/firebase-js-sdk/blob/5ad7ff2ae955c297556223e6cb3ad9d4b897f664/packages/util/src/errors.ts#L74

jhuleatt: or FirebaseError?

src/auth.tsx
jhuleatt: made it empty

src/auth.tsx
jhuleatt: made ReactFireError

src/auth.tsx
jhuleatt: added a todo inline

src/auth.tsx
```diff
@@ -59,9 +60,153 @@ export interface ClaimsCheckProps {
   user: firebase.User;
   fallback: React.ReactNode;
   children: React.ReactNode;
-  requiredClaims?: { [key: string]: any };
+  requiredClaims: { [key: string]: any };
 }
 
+interface ClaimCheckErrors {
+  [key: string]: ReactFireError[];
```
jamesdaniels, line 67: Do we want to restrict developers to ReactFireError for custom validators? Or should we do any[]?

jamesdaniels (approved):
LGTM, though see comment.

src/auth.tsx
jhuleatt: good point, made this an `any[]`

MrHazimAli:
hye @jhuleatt , may I know if there is any eta on new version include this PR? 👍  looking forward to it :-)

jhuleatt:
Hi @MrHazimAli, we're waiting on #372 before publishing a new release candidate, but in the meantime you can try out the new hooks with a canary version: `npm install reactfire@canary`

PR #370, opened 2021-05-04
Set `hasEmitted` for `initialData`

### LINKED ISSUE #342: Observable hasEmitted with useUser Hook never gets true
**React:** 17.0.1

**Firebase:**8.2.9

**ReactFire:**3.0.0rc.0

### Test case

https://codesandbox.io/s/musing-sky-dt1zi?file=/src/App.js

### Steps to reproduce

Use useUser in a component and load the page for first time. Try to use a loading state with
`if (status === "loading" || hasEmitted === false)` and see that hasEmitted never gets true, so the loading state never disappears.

Note: When you use hot reloading, all rerenders are not affected by this bug.

### Expected behavior
hasEmitted should be true after some time.

### Actual behavior

hasEmitted is always false.

jhuleatt:
### Description

Fixes #342 

- Have `useObservable` set `hasEmitted` to `true` when `initialData` is passed in
    - Added a test to verify this behavior
- Update `ObservableStatus` docs to make the values more clear

### DIFF
--- a/docs/reference/interfaces/useobservable.observablestatus.md
+++ b/docs/reference/interfaces/useobservable.observablestatus.md
@@ -27,7 +27,11 @@
 
 • **data**: T
 
-Defined in: [src/useObservable.ts:36](https://github.com/FirebaseExtended/reactfire/blob/main/src/useObservable.ts#L36)
+The most recent value.
+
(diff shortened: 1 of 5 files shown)


jamesdaniels (approved):
LGTM

PR #372, opened 2021-05-14
Update rxfire and rxjs

jhuleatt:


jhuleatt:
@jamesdaniels TS is angry, and I think it's due to the `add` hack you mentioned the other day. Can you please take a look?

Screen Shot 2021-05-14 at 12 58 45 PM

jamesdaniels:
@jhuleatt you can't chain off `add` any-longer.

This should work:
```ts
this._innerSubscriber.add(this._reset);
return this._innerSubscriber;
```

jhuleatt:
> @jhuleatt you can't chain off `add` any-longer.
> 
> This should work:

Thanks! That did. Just investigating a failing test and then this should be ready for review

jamesdaniels (approved):

package.json
```diff
@@ -86,18 +87,18 @@
...
     "typedoc-plugin-markdown": "^3.6.1",
-    "typescript": "^4.0.3"
+    "typescript": "^4.2.4"
   },
   "dependencies": {
-    "babel-plugin-minify-replace": "^0.5.0",
-    "rxfire": "^4.0.0",
-    "rxjs": "^6.6.3"
+    "rxfire": "5.0.0-rc.0",
+    "rxjs": "^7.0.1"
```
jamesdaniels, line 97: We should probably keep 6 while feasible, since 6 => 7 is a big jump for developers.

E.g, `^6.6.3 || ^7.0.1`

We could add to the matrix test like we did RxFire. WDYT?

src/sdk.tsx
```diff
@@ -112,7 +114,7 @@ function preloadFactory(componentName: 'performance'): (options: PreloadOptions<
 function preloadFactory(componentName: 'remoteConfig'): (options: PreloadOptions<App['remoteConfig']>) => Promise<App['remoteConfig']>;
 function preloadFactory(componentName: 'storage'): (options: PreloadOptions<App['storage']>) => Promise<App['storage']>;
 function preloadFactory(componentName: ComponentName) {
-  return (options: PreloadOptions<FirebaseInstanceFactory>) => preload(componentName, options.firebaseApp, options.setup).toPromise();
+  return (options: PreloadOptions<FirebaseInstanceFactory>) => firstValueFrom(preload(componentName, options.firebaseApp, options.setup));
```
jhuleatt, line 117: because [`toPromise` is deprecated](https://rxjs.dev/deprecations/to-promise)

src/SuspenseSubject.ts
```diff
@@ -84,7 +84,8 @@ export class SuspenseSubject<T> extends Subject<T> {
       clearTimeout(this._timeoutHandler);
     }
     this._innerSubscriber = this._innerObservable.subscribe(subscriber);
-    return this._innerSubscriber.add(this._reset);
+    // this._innerSubscriber.add(this._reset.bind(this));
```
jhuleatt, line 87: @jamesdaniels this was causing useSDK to throw out its value after it loaded. Would it be bad if I removed it?

jhuleatt: This could be important for other functions like `useUser`

Suggestion from James: have a flag `resetOnUnsubscribe` that's true for everything except the `useSDK` functions

package.json
jhuleatt: Convo with James: keep 6 in for now to avoid version conflicts if someone is using 6 in their app

src/sdk.tsx
jhuleatt: note: this doesn't exist in rxjs 6

jhuleatt: TODO: when we have a major that drops rxjs 6 we can switch, for now we can just use `toPromise`

package.json
jhuleatt: done

src/sdk.tsx
jhuleatt: done

src/SuspenseSubject.ts
jhuleatt: added a test for this, and it passes without adding `reset`, so I'm removing that line

jhuleatt:
I think this is blocked by https://github.com/FirebaseExtended/rxfire/issues/13

I can't get ReactFire's example web app working because of that issue, and I'd really like to click around in the example for peace of mind before merging

jhuleatt:
> I think this is blocked by [FirebaseExtended/rxfire#13](https://github.com/FirebaseExtended/rxfire/issues/13)
> 
> I can't get ReactFire's example web app working because of that issue, and I'd really like to click around in the example for peace of mind before merging

I've fixed the rxfire issue and released `rxfire@5.0.0-rc.1`. With that build, now all the tests pass AND I'm able to click around in the example app! Everything looks good.

PR #392, opened 2021-06-27
Update readme for concurrent rendering

webbertakken:
React "Concurrent mode" is now "concurrent rendering".

We might also refer to the new documentation [blog page](https://reactjs.org/blog/2021/06/08/the-plan-for-react-18.html) for react 18 but it is much more generic, so perhaps keep it as is for now.

Unless I understand `ObservableStatus<T>` incorrectly somehow, we need to destructure `data: T`. I updated the doc accordingly.


webbertakken:
@googlebot I signed it!

jhuleatt:
Thank you @webbertakken!

PR #410, opened 2021-07-26
Align validateClaims to validateCustomClaims

banyan:
When I passed `validateCustomClaims`, the callback was not called, so I looked at the code and found that [the key name was different](https://github.com/FirebaseExtended/reactfire/blob/804b3e56ccc3d1595431437627f9baeb362c4d37/src/auth.tsx#L162) from what I was expecting.


jhuleatt (approved):

jhuleatt:
Nice catch @banyan, thank you for fixing this and even adding a test! 🌟

banyan:
@jhuleatt 

Thanks for the quick review and also the release!

PR #413, opened 2021-07-27
[firebase v9] Add useInitSdk functions

jhuleatt:
### Description

Make it easier to initialize a Firebase SDK in the new modular SDK

Rules for the `useInit*` hooks:

- Must be called inside `FirebaseAppProvider`
- Cannot be called inside their SDK provider (For example, `useInitAuth` can't be called inside `AuthProvider`)

#### `useInitFirestore`

```jsx
// useInitFirestore is new!
import { useInitFirestore, FirestoreProvider } from "reactfire";

function Comment({ commentId }) {
  // initialize Firestore
  // could even do an async import of Firestore in the callback here
  const { status, data: firestoreInstance } = useInitFirestore(
    async firebaseApp => {
      const db = initializeFirestore(firebaseApp, {});
      await enableIndexedDbPersistence(db);
      return db;
    }
  );

  // firestore init isn't complete yet
  if (status === "loading") {
    return <LoadingSpinner />;
  }

  // pass the Firestore instance to FirestoreProvider
  // now we can be sure that any child of FirestoreProvider
  // has a fully initialized Firestore instance with
  // indexedDbPersistence enabled
  return (
    <FirestoreProvider sdk={firestoreInstance}>
      <CommentText commentId={commentId} />
      <LikeCount commentId={commentId} />
    </FirestoreProvider>
  );
}
```

#### `useInitRemoteConfig`

```jsx
// useInitRemoteConfig is new!
const { status, data: remoteConfigInstance } = useInitRemoteConfig(
  async firebaseApp => {
    const remoteConfig = getRemoteConfig(firebaseApp); // this is the normal getRemoteConfig from the Firebase SDK
    remoteConfig.settings = {
      minimumFetchIntervalMillis: 10000,
      fetchTimeoutMillis: 10000
    };
    // now we can be sure RC is fully activated before anyone tries to use it
    await fetchAndActivate(remoteConfig);
    return remoteConfig;
  }
);

if (status === "loading") {
  return <LoadingSpinner />;
}

return (
  <RemoteConfigProvider sdk={remoteConfigInstance}>
    <RcString messageKey="message" />
  </RemoteConfigProvider>
);
```

#### `useInitPerformance`

It also provides a convenient way to initialize performance monitoring asynchronously

```jsx
useInitPerformance(async firebaseApp => {
  const { getPerformance } = await import("firebase/performance");
  return getPerformance(firebaseApp);
});
```

### These are all async. What if I want a synchronous solution?

In that case, you can just initialize things with the Firebase JS SDK itself:

```jsx
const firebaseApp = useFirebaseApp();
const auth = getAuth(firebaseApp);

return (<AuthProvider sdk={auth}>{children}</AuthProvider>)
```

### DIFF
--- a/example/withoutSuspense/Firestore.tsx
+++ b/example/withoutSuspense/Firestore.tsx
@@ -1,12 +1,11 @@
 import * as React from 'react';
-import { useEffect, useState } from 'react';
-import { FirestoreProvider, useFirebaseApp, useFirestore, useFirestoreCollectionData, useFirestoreDocData, useFirestoreDocDataOnce } from 'reactfire';
+import { useState } from 'react';
+import { FirestoreProvider, useFirebaseApp, useFirestore, useFirestoreCollectionData, useFirestoreDocData, useFirestoreDocDataOnce, useInitFirestore } from 'reactfire';
 import { WideButton } from '../display/Button';
 import { CardSection } from '../display/Card';
 import { LoadingSpinner } from '../display/LoadingSpinner';
 import { AuthWrapper } from './Auth';
 import { initializeFirestore, doc, collection, enableIndexedDbPersistence, increment, updateDoc, orderBy, query, addDoc } from 'firebase/firestore';
-import type { FirebaseFirestore } from 'firebase/firestore';
 import { getFirestore } from 'firebase/firestore';
 
 const Counter = () => {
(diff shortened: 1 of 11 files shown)


example/withoutSuspense/Firestore.tsx
```diff
@@ -113,22 +112,13 @@ const StaticValue = () => {
...
-  const firebaseApp = useFirebaseApp();
-  const [firestoreInstance, setFirestoreInstance] = useState<FirebaseFirestore | undefined>();
-
-  // idea: useFirestoreInstance();
-  // this could set up enableIndexedDbPersistence
-  // and emulators!
-  // https://github.com/angular/angularfire/issues/1158
-
-  useEffect(() => {
+  const {status, data: firestoreInstance} = useInitFirestore(async (firebaseApp) => {
```
davideast, line 115: This looks really good for custom initialization. If you want to support lazy-loading, we'll need to have an API that stores the entire module so it's easier to retrieve later without having to async fetch it again.

```ts
const { status, data: firestoreInstance } = useInitFirestore(
    async firebaseApp => {
       // It does this behind this scenes -> import('firebase/firestore');
       // and then sets to a provider so you can import it later.
      const firestoreModule = await useLazyFirestore();
      const db = firestoreModule.initializeFirestore(firebaseApp, { });
      await firestoreModule.enableIndexedDbPersistence(db);
      return db;
    }
  );
```

That' another API for another day. But it's good to consider it for this design.

example/withoutSuspense/RemoteConfig.tsx
```diff
@@ -19,20 +19,23 @@ export const RcString = ({ messageKey }) => {
 };
 
 export const RemoteConfig = () => {
-  const firebaseApp = useFirebaseApp();
-  const remoteConfig = getRemoteConfig(firebaseApp);
-  const [remoteConfigLoaded, setRemoteConfigLoaded] = useState<boolean>(false);
-  useEffect(() => {
+  const { status, data: remoteConfigInstance } = useInitRemoteConfig(async (firebaseApp) => {
```
davideast, line 22: This is really nice for RC!

davideast (approved):
Looks great! Let's merge it.

PR #435, opened 2021-08-27
Resolve or Replace @ts-ignore

sujishpatel:
### Description
1. Reordered error checking for SdkProvider so that it will actually throw error for `no sdk provided`.
2. Resolve @ts-ignore in SdkProvider: v9 exposes props.sdk.app.name for all SDKs, so used that for app name comparison.
3. For the remaining ts-ignores, replaced with @ts-expect-error.

### DIFF
--- a/docs/reference/modules/sdk.md
+++ b/docs/reference/modules/sdk.md
@@ -46,7 +46,7 @@
 
 #### Defined in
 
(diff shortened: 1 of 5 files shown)


jhuleatt:
Thanks for this cleanup @sujishpatel! Can you please run `yarn docs` from the root of the project to make that last check go green?

jhuleatt (approved):

jhuleatt:
Thank you @sujishpatel!

PR #444, opened 2021-09-03
Add support for Cloud Functions for Firebase

### LINKED ISSUE #330: Preload functions to set region
Hi,

I am trying to preload functions to set their region.

### Version info

**React:** 17.0

**Firebase:** 8.2.4

**ReactFire:** 3.0.0

### Steps to reproduce

In my index file, I am preloading firebase components as stated in the guides, but trying to preload functions, too. When I try to use functions in another component, the set region defaults to us-central1 instead of using the custom region.

```
    preloadFunctions({ 
      firebaseApp, 
      setup: (factory) => {
        
      const functions = factory("europe-west1")
      console.log("Function Preload", functions)
      return functions ;
    }
```
console.log returns function with region "europe-west1"

However in another react component that is called afterwards I use
```
const OtherComponent =  () => {
  const functions = useFunctions();
  console.log("Function Load ", functions)
...}
```
This console.log returns "us-central1" as region

### Expected behavior

Both functions should return in their console.logs region "europe-west1"

### Actual behavior

Preloaded function returns console.logs region "europe-west1", the useFunction function however returns "us-central1"

Further on, it seems the setup of preloadFunctions expects a Promise as a return value, however functions is not a Promise but of type firebase.functions.Functions. 
It would be the same i.e. when preloading firestore. The setup would complain, if one would return firestore(), but it works i.e. for firestore().enablePersistence(), since it is a Promise. Functions however won't return a Promise.

```
preloadFirestore({app, setup(firestore){ return firestore()}) // has type error, since firestore() is not a Promise
preloadFirestore({app, setup(firestore){ return firestore().enablePersistence()}) // has no type error, since it returns Promise
preloadFunction({app, setup(function){ return function("us-central1")}) // has type error, since firestore() is not a Promise
preloadFunction({app, setup(function){ return}) // has no type error, since return is void
```

My guess is, that functions cannot be preloaded, maybe except for using useEmulator? How would I set the function region though?
`useFunctions` does not seem to support it, but this seems to work currently:
````
  const app = useFirebaseApp()
  const functions = app.functions("europe-west1");
  const myFunction= functions.httpsCallable("myFunction")
```

Kind regards!

jhuleatt:
### Code sample

```jsx
function Calculator() {
  const functions = useFunctions();
  const remoteCalculator = httpsCallable(functions, 'calculate');

  const [calculationResult, setResult] = useState(null);

  async function handleButtonClick(firstNumber, secondNumber, operator) {
    const remoteCalculatorResponse = await remoteCalculator({ firstNumber, secondNumber, operator });
    setResult(remoteCalculatorResponse.data);
  }

  if (!calculationResult) {
    return <button onClick={() => handleButtonClick(1, 2, '+')}>Click me to add 1 + 2</button>;
  } else {
    return <pre>{calculationResult}</pre>;
  }
}
```

---

fixes #330

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -95,7 +95,7 @@ jobs:
           yarn install --frozen-lockfile --prefer-offline
       - name: Install deps for functions
         run: /
-          (cd functions && npm install)
+          cd functions ; npm install ; cd ..
       - name: Firebase emulator cache
         uses: actions/cache@v2
         with:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -93,6 +93,9 @@ jobs:
         run: |
           yarn config set yarn-offline-mirror ~/.npm-packages-offline-cache
           yarn install --frozen-lockfile --prefer-offline
+      - name: Install deps for functions
+        run: /
+          (cd functions && npm install)
       - name: Firebase emulator cache
         uses: actions/cache@v2
         with:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -94,7 +94,7 @@ jobs:
           yarn config set yarn-offline-mirror ~/.npm-packages-offline-cache
           yarn install --frozen-lockfile --prefer-offline
       - name: Install deps for functions
-        run: /
+        run: |
           cd functions ; npm install ; cd ..
       - name: Firebase emulator cache
         uses: actions/cache@v2
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -94,8 +94,8 @@ jobs:
           yarn config set yarn-offline-mirror ~/.npm-packages-offline-cache
           yarn install --frozen-lockfile --prefer-offline
       - name: Install deps for functions
-        run: |
-          cd functions ; npm install ; cd ..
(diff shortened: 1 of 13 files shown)


jhuleatt:
Tests are failing because I need to update the workflow yaml to run `npm install` in the functions directory

sujishpatel:
I was considering sending a PR for this, then thought that it might be nice to create some more useful helpers around the functions library.

The following are a few thoughts
```
type HttpsCallableOptions = {
   suspense: boolean,
}
const [send, data, loading, ] = useHttpsCallable('calculate', {}:Options);
```

sujishpatel (approved):
LGTM

.github/workflows/test.yaml
```diff
@@ -93,6 +93,9 @@ jobs:
         run: |
           yarn config set yarn-offline-mirror ~/.npm-packages-offline-cache
           yarn install --frozen-lockfile --prefer-offline
+      - name: Install deps for functions
+        run: |
```
jhuleatt, line 97: can probably specify the directory for this step

jamesdaniels (approved):

.github/workflows/test.yaml
jhuleatt: done

PR #446, opened 2021-09-07
Make sure Firestore hooks return `undefined` when data does not exist

### LINKED ISSUE #440: usesFirestoreDocData stuck in `status: "loading"` if data does not exist
_Originally posted by @namiwang in https://github.com/FirebaseExtended/reactfire/discussions/438#discussioncomment-1251839_

### Expected behavior

```js
{
  // ...
  status: "success",
  data: undefined,
  // ...
}
```

### Actual behavior

```js
{
  // ...
  status: "loading",
  data: undefined,
  // ...
}
```

jhuleatt:
fixes #440 

- [x] repro with a failing test

### DIFF
--- a/src/useObservable.ts
+++ b/src/useObservable.ts
@@ -63,55 +63,75 @@ export interface ObservableStatus<T> {
   firstValuePromise: Promise<void>;
 }
 
+function reducerFactory<T>(observable: SuspenseSubject<T>) {
+  return function reducer(state: ObservableStatus<T>, action: 'value' | 'error' | 'complete'): ObservableStatus<T> {
+    // always make sure these values are in sync with the observable
+    const newState = {
+      ...state,
+      hasEmitted: state.hasEmitted || observable.hasValue,
+      error: observable.ourError,
+      firstValuePromise: observable.firstEmission,
+    };
+    if (observable.hasValue) {
+      newState.data = observable.value;
+    }
+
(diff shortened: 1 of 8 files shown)


jhuleatt:
Turns out this is an issue with `useObservable` itself.

src/useObservable.ts
```diff
@@ -63,55 +63,75 @@ export interface ObservableStatus<T> {
...
+    data: observable.hasValue ? observable.value : config?.initialData ?? config?.startWithValue,
+    error: observable.ourError,
+    firstValuePromise: observable.firstEmission,
+  };
+  const [status, dispatch] = React.useReducer<React.Reducer<ObservableStatus<T>, 'value' | 'error' | 'complete'>>(reducerFactory<T>(observable), initialState);
+
   React.useEffect(() => {
     const subscription = observable.subscribe({
-      next: (v) => {
-        setValue(() => v);
```
jhuleatt, line 86: This is where the original issue was. Since `value` was undefined when the state was set up, calling `setValue(() => undefined)` meant that React didn't trigger a re-render, since `value` didn't change.

jhuleatt:
I'm going to admin-merge this because #438 is a pretty active discussion so a fast fix would be beneficial, I was able to repro the original issue with a test that is now green, and `useObservable` has pretty solid test coverage.

PR #449, opened 2021-09-10
add useCallableFunctionResponse

### LINKED ISSUE #447: Feature Request/Proposal: Expanding Functions
A few thoughts on expanding the Functions API.

I was thinking if it would be worthwhile to expand the functions API a bit, mainly in regard to handling loading and error states. 
And potentially expose some helpful options.  (I haven't fully thought through the various options, but wanted to start a discussion on them)

```ts
// Sorta risk combining types like this, but can't think of a cleaner way to keep things simple. 
type ReactFireHttpsCallableOptions & HttpsCallableOptions = {
   suspense: boolean, // Enable suspense? Maybe just use the internal settings to determine if we should suspend.
   start: boolean, // Calls the functions right away. 
   data?: RequestData // Firebase Typing, To use in conjunction with start option
   warmup: boolean // See Below
}
type Status = {
loading: boolean
error: FunctionsError // Firebase Typing
}
const [data, send, status] = useHttpsCallable<Req,Res>('calculate', {}:ReactFireHttpsCallableOptions) ;

```

## ReactFireHttpsCallableOptions Breakdown
* warmup: I know cold boot can be a massive pain point for some. To help alleviate this problem, I was thinking ReactFire can just make an OPTIONS request when `useHttpsCallable` is called. Since httpsCallable functions are just POST endpoints, this should be safe. Recreating the code to determine the url should be fairly trivial (just need to take into account; custom domains and region)
* timeout: Comes from firebase's HttpsCallableOptions
* start/data/suspense: Not fully sold on these, just wanted to mention them

## useHttpsCallable / useFunctionsCallable / useFunctionsHttpsCallable 
* data: The payload of the function
* send: Invokes the function. Optionally takes in ResponseData
* status: An object with two properties, loading: boolean and  error: FunctionsError | undefined

## Example Use Cases
Simple
```tsx
const [data, send, {loading, error} ] = useHttpsCallable('MyFunctions');

if (loading) return <LoadingComponent/>
if (error) return <ErrorFallback {...error}/>
return <button onClick={()=> send('my payload')}> Start </button>
```
Call function right away. 
```tsx
const [data, send, {loading, error} ] = useHttpsCallable('MyFunctions', { start: true, data: props.id });

if (loading) return <LoadingComponent/>
if (error) return <ErrorFallback {...error}/>
return <div>{data.message}</div>
```

jhuleatt:
### Description

Add the `useCallableFunctionResponse` hook, which makes it easy to call a callable function on render.

fixes #447 

### Code sample

```jsx
import { useCallableFunctionResponse } from "reactfire";

function LikeCount({ videoId }) {
  const { status, data: likeCount } = useCallableFunctionResponse(
    "countVideoLikes",
    { data: { videoId: videoId } }
  );

  return (
    <span>This video has {status === "loading" ? "..." : likeCount} likes</span>
  );
}
```

---

Note: We also discuss `useGetCallableFunction` in #447, but I've left that out of this PR. After some more thought, I don't think it is worth the maintenance burden, since it is easy enough to use the Firebase JS SDK directly for that use case:

```js
import { useFunctions } from "reactfire";
import { httpsCallable } from "firebase/functions";

function MyComponent() {
  const remoteCalculator = httpsCallable(useFunctions(), "calculate");

  //...
}
```

### DIFF
--- a/docs/use.md
+++ b/docs/use.md
@@ -14,7 +14,8 @@
   * [Show a single document](#show-a-single-document)
   * [Show a list of data (collection)](#show-a-list-of-data-collection)
 - [Cloud Functions](#cloud-functions)
-  * [Call a function](#call-a-function)
+  * [Call a function based on user interaction](#call-a-function-based-on-user-interaction)
+  * [Call a function on render](#call-a-function-on-render)
 - [Realtime Database](#realtime-database)
   * [Show an object](#show-an-object)
   * [Show a list of data](#show-a-list-of-data)
@@ -304,7 +305,7 @@ function FavoriteAnimals() {
 
 The following samples assume that `FirebaseAppProvider` and `FunctionsProvider` components exist higher up the component tree.
 
-### Call a function
+### Call a function based on user interaction
 
 ```jsx
 function Calculator() {
@@ -324,6 +325,18 @@ function Calculator() {
   } else {
     return <pre>{calculationResult}</pre>;
   }
+}
+```
+
+### Call a function on render
+
+If you want to call a function when a component renders, instead of in response to user interaction, you can use the `useCallableFunctionResponse` hook.
+
+```jsx
+function LikeCount({ videoId }) {
+  const { status, data: likeCount } = useCallableFunctionResponse('capitalizeText', { data: { videoId: videoId } });
+
+  return <span>This video has {status === 'loading' ? '...' : likeCount} views</span>;
 }
 ```
 
--- a/docs/use.md
+++ b/docs/use.md
@@ -336,7 +336,7 @@ If you want to call a function when a component renders, instead of in response 
 function LikeCount({ videoId }) {
   const { status, data: likeCount } = useCallableFunctionResponse('countVideoLikes', { data: { videoId: videoId } });
 
-  return <span>This video has {status === 'loading' ? '...' : likeCount} views</span>;
+  return <span>This video has {status === 'loading' ? '...' : likeCount} likes</span>;
 }
 ```
 
--- a/docs/use.md
+++ b/docs/use.md
@@ -334,7 +334,7 @@ If you want to call a function when a component renders, instead of in response 
 
 ```jsx
 function LikeCount({ videoId }) {
-  const { status, data: likeCount } = useCallableFunctionResponse('capitalizeText', { data: { videoId: videoId } });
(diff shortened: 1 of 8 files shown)


src/functions.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import type { HttpsCallableOptions } from 'firebase/functions';
+
+/**
+ * Calls a callable function.
+ *
+ * @param functionName - The name of the function to call
+ * @param options
+ */
+export function useCallableFunctionResponse<RequestData, ResponseData>(
```
jhuleatt, line 13: ☝️ this is where it is implemented

src/functions.tsx
```diff
@@ -0,0 +1,28 @@
...
+  options?: ReactFireOptions<ResponseData> & {
+    httpsCallableOptions?: HttpsCallableOptions;
+    data?: RequestData;
+  }
+): ObservableStatus<ResponseData> {
+  const functions = useFunctions();
+  const observableId = `functions:callableResponse:${functionName}:${JSON.stringify(options?.data)}:${JSON.stringify(options?.httpsCallableOptions)}`;
+  const obsFactory = rxHttpsCallable<RequestData, ResponseData>(functions, functionName, options?.httpsCallableOptions);
+
+  //@ts-expect-error because RxFire doesn't make data optional. Remove when https://github.com/FirebaseExtended/rxfire/pull/34 is released.
```
jhuleatt, line 24: https://github.com/FirebaseExtended/rxfire/pull/34

docs/use.md
```diff
@@ -327,6 +328,18 @@ function Calculator() {
 }
 ```
 
+### Call a function on render
+
+If you want to call a function when a component renders, instead of in response to user interaction, you can use the `useCallableFunctionResponse` hook.
+
+```jsx
+function LikeCount({ videoId }) {
+  const { status, data: likeCount } = useCallableFunctionResponse('countVideoLikes', { data: { videoId: videoId } });
```
jhuleatt, line 337: What do you think about the name `useCallableFunctionResponse`? It's a little wordy because I wanted to emphasize that this would call the callable function and update with the response. Any suggestions to make it shorter/better?

jamesdaniels: Wordy but I like that it's very explicit. SGTM

jamesdaniels (approved):
LGTM

PR #486, opened 2021-11-11
Fix reconciler : Expose the contexts

RenaudRohlinger:
Hello! 

### Description

To consume the contexts with [react reconciler](https://www.npmjs.com/package/react-reconciler) you need to forward the contexts using a "bridge".
See https://github.com/facebook/react/issues/17275
Exposing each context is then necessary if we don't want to wrap all our components twice in that case.

### DIFF
--- a/src/sdk.tsx
+++ b/src/sdk.tsx
@@ -15,15 +15,15 @@ import { ObservableStatus, useObservable } from './useObservable';
 import { from } from 'rxjs';
 import { ReactFireOptions } from '.';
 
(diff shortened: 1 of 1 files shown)


RenaudRohlinger:
@googlebot I signed it!

jhuleatt:
Hi @RenaudRohlinger, thanks for the PR! While this seems like a rare use case (react-reconciler is "Use it at your own risk" according to the React docs), I think it should be fine to merge if it will benefit you.

RenaudRohlinger:
Thanks!

PR #505, opened 2022-02-05
updating docs from 'useSignInCheck' to 'useSigninCheck'

thomasmburke:
Ran into import errors when following the docs because of this subtle lowercase `i`.

I think overall it would be worth having this camel case exception be consistent with `signInCheckResult`, but not worth introducing a breaking change in my opinion, so just looking to update the docs and references in this repo.

### DIFF
--- a/docs/reference/modules/auth.md
+++ b/docs/reference/modules/auth.md
@@ -43,7 +43,7 @@
 
 ▸ **AuthCheck**(`__namedParameters`): `JSX.Element`
 
-**`deprecated`** Use `useSignInCheck` instead
+**`deprecated`** Use `useSigninCheck` instead
 
 Conditionally render children based on signed-in status and [custom claims](https://firebase.google.com/docs/auth/admin/custom-claims).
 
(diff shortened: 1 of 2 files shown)


jhuleatt (approved):

jhuleatt:
Thanks @thomasmburke!

jhuleatt (changes requested):
Thanks! Can you please update in the source as well so that these changes persist next time the docs are generated?

https://github.com/FirebaseExtended/reactfire/blob/eb1428b9e7923a898cff73ca4725c012b4074bae/src/auth.tsx#L204

https://github.com/FirebaseExtended/reactfire/blob/eb1428b9e7923a898cff73ca4725c012b4074bae/src/auth.tsx#L241

jhuleatt:
fixed in this commit: https://github.com/FirebaseExtended/reactfire/commit/414201e88bd6f6701998993f6203e4d49773147b

thomasmburke:
Sorry for missing this!

Issue #495, opened 2021-12-29
useUser() status is "success" on first render despite the value being undefined


ralrom:
### Version info

**Next:** 12.0.7
**React:** 17.0.2
**Firebase:** ^9.6.1
**ReactFire:** ^4.2.1
**Suspense mode**: off

### Test case

```tsx
// FirebaseProviders.tsx

const firebaseConfig = {
  // ...
}

function FirebaseServices({ children }) {
  const app = useFirebaseApp();
  const firestore = getFirestore(app);
  const auth = getAuth(app);
  const storage = getStorage(app);

  return (
    <FirestoreProvider sdk={firestore}>
      <StorageProvider sdk={storage}>
        <AuthProvider sdk={auth}>{children}</AuthProvider>
      </StorageProvider>
    </FirestoreProvider>
  );
}

export default function FirebaseProviders({ children }) {
  return (
    <FirebaseAppProvider firebaseConfig={firebaseConfig}>
      <FirebaseServices>{children}</FirebaseServices>
    </FirebaseAppProvider>
  );
}
```

```tsx
// MainMenu.tsx

export default function MainMenu() {
  const { status, data: user } = useUser();
  console.log("status", status, "user", user);
  return <div>test</div>
}
```

### Expected Output

```
status loading user undefined (first render)
status success user null (2nd render)
```

I expect status === "loading" for the first render

### Actual behavior

```
status success user undefined (first render)
status success user null (2nd render)
```

I get status === "success" for the first render despite the value still being undefined at that point

TatiSur:
have the same :(
any solutions?

MatthewFallon:
Found one possible fix to apply for this. The useUser method is creating the observable with useObservable and passing it a initialData value of `null`, I assume with the expectation that that is equivalent to not passing an initial data value... BUT in useObservable the method is using Object.hasOwnProperty to check for the existence of the initialData prop and hasOwnProperty doesn't check for the falsy value of a property it checks for existence.

Quote code from MDN [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#using_hasownproperty_to_test_for_an_own_propertys_existence):
```
let example = {};
example.hasOwnProperty('prop');   // false

example.prop = 'exists';
example.hasOwnProperty('prop');   // true - 'prop' has been defined

example.prop = null;
example.hasOwnProperty('prop');   // true - own property exists with value of null

example.prop = undefined;
[...]

PR #507, opened 2022-02-19
Fix for evaluating initialData of observable

MatthewFallon:
Changing the useObservable method to
evaluate initialData as falsy rather than
using hasOwnProperty which accidentally
returns "success" early from an observable.

### Description

CLA signed for this account.

Relevant issue: #495 

Tests for useObservable still passing. I did not get the emulators set up properly for some of the additional tests on my own machine. 

Currently your docs state that useUser should be returning "loading" until it emits it's first value, however you are accidentally interpreting `initialData: undefined | null` as equal to `true` thus returning the wrong status on first render. This fixes that. I simply changed one line to make it so that it evaluates the falsy value of the initialData coming in rather than the existence of the property, so that it does not have false positives like this in the future. I am not certain if this could cause other unintended consequences, I would appreciate if a maintainer has time to take a look.

### DIFF
--- a/src/useObservable.ts
+++ b/src/useObservable.ts
@@ -100,7 +100,9 @@ export function useObservable<T = unknown>(observableId: string, source: Observa
   const observable = preloadObservable(source, observableId);
 
   // Suspend if suspense is enabled and no initial data exists
-  const hasInitialData = config.hasOwnProperty('initialData') || config.hasOwnProperty('startWithValue');
+  // Changed from using Object.hasOwnProperty() to evaluating
+  // the falsy of the properties as a boolean.
+  const hasInitialData = !!(config.initialData || config.startWithValue);
   const hasData = observable.hasValue || hasInitialData;
   const suspenseEnabled = useSuspenseEnabledFromConfigAndContext(config.suspense);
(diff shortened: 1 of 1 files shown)


jhuleatt:
@MatthewFallon sorry we were slow to respond here.

`initialData`'s type is `any`:

https://github.com/FirebaseExtended/reactfire/blob/eb1428b9e7923a898cff73ca4725c012b4074bae/src/index.ts#L26

Because of this, `undefined` and `null` technically are valid values. `null` is especially important, because that's what the Firebase Auth SDK returns from [`auth.currentUser` ](https://firebase.google.com/docs/reference/js/auth.auth.md#authcurrentuser) when someone isn't signed in.

We could consider ignoring `undefined` for `initialData`, but that would be a breaking change that would require us to change the type of `initialData`. If it's something you feel strongly about and would like to create an issue to discuss for ReactFire v5, we can definitely consider it! Otherwise we should stick with the fix in #521.

MatthewFallon:
Ok, that definitely makes sense, I didn't even consider the impact it would have on if a user wasn't signed in at all for some reason. Thank you for dropping a reply I appreciate it. :+1:

PR #525, opened 2022-06-10
Disable observable timeouts for suspense

Mythie:
### Description

When using reactfire with react 17, any request that exceeds the default timeout of 30 seconds will then trigger the timeout handler that resets the observable. 

Unfortunately, this resetting of the observable also means that the `firstEmission` promise never gets resolved leading to an infinite fallback state with suspense.

To resolve this, we pass along whether or not we are in suspense mode when creating an observable so we may conditionally choose to allow the timeout for resetting the observable.


jhuleatt (approved):
This is a really interesting fix, thank you @Mythie! Given the number of thumbs-ups on this PR and that the tests pass, I'm going to merge it

PR #527, opened 2022-07-01
pass options from useSigninCheck to useObservable

CapsAdmin:
### Description

While trying to debug while useSigninCheck suspends indefinitely on latest version and master, I tried to disable suspense but this was not working. Having cloned the repo locally and debugged I noticed that the options used in useSigninCheck was never passed to useObservable which takes the options object.

### DIFF
--- a/src/auth.tsx
+++ b/src/auth.tsx
@@ -180,7 +180,7 @@ export function useSigninCheck(
     })
   );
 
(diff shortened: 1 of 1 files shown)


jhuleatt:
A very belated thank you, @CapsAdmin!

PR #534, opened 2022-08-03
Move build from TSDX to Vite

jhuleatt:
- Pulls out TSDX, replaces with Vite
- Updates dependencies
- Moves from yarn to npm now that we don't need yarn workspaces
- Uses Vitest instead of Jest
- Updates Firebase CLI to v11, which requires removing Node 12 from our tests since the CLI requires Node 14+
- Running the build locally creates a `stats.html` file. Open that in the browser to inspect the built bundle


jhuleatt:
I'm not sure why the Storage tests keep failing only on Node 14. I can run them locally in Node 14 with no issue.

jamesdaniels (approved):

PR #589, opened 2023-08-03
JS v9.23

christhompsongoogle:
Upgrading to 9.23 to capture some multiDB API changes.

Removed some unused dependencies.


jhuleatt (approved):
Thanks @christhompsongoogle!

PR #640, opened 2025-03-25
NextJS middleware

jamesdaniels:

### DIFF
--- a/src/nextjs/index.ts
+++ b/src/nextjs/index.ts
@@ -1,244 +1,601 @@
-import { NextResponse, NextRequest } from 'next/server';
-import { createRemoteJWKSet, jwtVerify, JWTPayload, JWTHeaderParameters } from 'jose';
+import { NextResponse, NextRequest } from "next/server";
+import { LRUCache } from "lru-cache";
+import {
+  createRemoteJWKSet,
+  createLocalJWKSet,
+  jwtVerify,
+  JWTPayload,
+  JWTHeaderParameters,
+  decodeJwt,
+  decodeProtectedHeader,
+  JSONWebKeySet,
+} from "jose";
 import { getDefaultAppConfig } from "@firebase/util";
-import { FirebaseOptions } from 'firebase/app';
+import type { FirebaseOptions } from "firebase/app";
 
-type FirebaseJWTPayload = JWTPayload & { firebase?: { tenant?: string }};
+// FYI this code is supposed to be part of the Firebase SDKs soon
+// https://github.com/FirebaseExtended/reactfire/pull/640
 
-type RunMiddlewareOptions = { apiKey: string, projectId: string, emulatorHost: string|undefined, tenantId: string|undefined, authDomain: string|undefined };
-type RunMiddlewareResponse = [NextResponse]|[undefined, ((response: NextResponse) => NextResponse), FirebaseJWTPayload|undefined];
-async function runMiddleware(options: RunMiddlewareOptions, request: NextRequest): Promise<RunMiddlewareResponse> {
-    if (request.nextUrl.pathname.startsWith("/__/") && options.authDomain) {
-        const newURL = new URL(request.nextUrl);
-        newURL.host = options.authDomain;
-        newURL.port = '';
-        return [NextResponse.rewrite(newURL)];
+export type FirebaseJWTPayload = JWTPayload & {
+  email?: string;
+  email_verified?: boolean;
+  picture?: string;
+  name?: string;
+  user_id?: string;
+  firebase?: { tenant?: string; [key: string]: unknown };
+  [key: string]: unknown;
+};
+
+export interface CacheProvider {
+  get<T = unknown>(key: string): Promise<T | null | undefined> | T | null | undefined;
+  set(
+    key: string,
+    value: any,
+    options?: { ex?: number; ttl?: number; nx?: boolean },
+  ): Promise<any> | any;
+  setex?(key: string, seconds: number, value: any): Promise<any> | any;
+}
+
+export class MemoryCacheProvider implements CacheProvider {
+  private cache = new LRUCache<string, any>({ max: 1000 });
+
+  get<T = unknown>(key: string): T | null {
+    return (this.cache.get(key) as T) ?? null;
+  }
+
+  set(key: string, value: any, options?: { ex?: number; ttl?: number; nx?: boolean }): any {
+    if (options?.nx && this.cache.has(key)) return null;
+    const ttlSeconds = options?.ex ?? (options?.ttl ? Math.floor(options.ttl / 1000) : undefined);
+    const ttlMs = ttlSeconds ? ttlSeconds * 1000 : undefined;
+    this.cache.set(key, value, { ttl: ttlMs });
+    return "OK";
+  }
+
+  setex(key: string, seconds: number, value: any): any {
+    return this.set(key, value, { ex: seconds });
+  }
+}
+
+const defaultMemoryCache = new MemoryCacheProvider();
+
+async function cacheGet<T>(cache: CacheProvider, key: string): Promise<T | null> {
+  try {
+    const val = await cache.get<T>(key);
+    return val ?? null;
+  } catch (e) {
+    console.error(`Cache get failed for key "${key}":`, e);
+    return null;
+  }
+}
+
+async function cacheSetEx(
+  cache: CacheProvider,
+  key: string,
+  ttlSeconds: number,
+  value: any,
+): Promise<void> {
+  try {
+    if (typeof cache.setex === "function") {
+      await cache.setex(key, ttlSeconds, value);
+    } else {
+      await cache.set(key, value, { ex: ttlSeconds });
+    }
+  } catch (e) {
+    console.error(`Cache set failed for key "${key}":`, e);
+  }
+}
+
+async function cacheSetNx(
+  cache: CacheProvider,
+  key: string,
+  ttlSeconds: number,
+  value: any,
+): Promise<boolean> {
+  try {
+    const res = await cache.set(key, value, { nx: true, ex: ttlSeconds });
+    return !!res;
+  } catch (e) {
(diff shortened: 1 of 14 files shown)


src/nextjs/index.ts
```diff
@@ -0,0 +1,199 @@
+import { NextResponse, NextRequest } from 'next/server';
+import { createRemoteJWKSet, jwtVerify, JWTPayload, JWTHeaderParameters } from 'jose';
+import { getDefaultAppConfig } from "@firebase/util";
+import { FirebaseOptions } from 'firebase/app';
+
+type FirebaseJWTPayload = JWTPayload & { firebase?: { tenant?: string }};
+
+type RunMiddlewareOptions = { apiKey: string, projectId: string, emulatorHost: string|undefined, tenantId: string|undefined, authDomain: string|undefined };
+type RunMiddlewareResponse = [NextResponse]|[undefined, ((response: NextResponse) => NextResponse), FirebaseJWTPayload|undefined];
+async function runMiddleware(options: RunMiddlewareOptions, request: NextRequest): Promise<RunMiddlewareResponse> {
```
jamesdaniels: Feedback from bugbash so far

1. logout should be handled by a third cookie
2. cookie name needs to not have the api key
3. idToken should have expire with the JWT
4. if no idToken, invalid, or expired use the refresh token—rather than trigger logout

jhuleatt:
Thanks James! @tyler-reitz please keep this PR open. This is a prototype Next.js middleware for Firebase Auth to allow sync-ing auth state between Server Components and Client Components.

tyler-reitz (changes requested):
Thanks for this PR, @jamesdaniels. A few things to address before merge:

**Blockers**

**jose in devDependencies (package.json line 26)**
`jose` is imported at runtime but listed under `devDependencies`. It won't be installed in production. Based on the README install command (`npm install firebase-cookie-middleware jose lru-cache`), the intent looks like `peerDependencies`, but that should be explicit. `lru-cache` is in `dependencies` (correct); `jose` should be too, or explicitly peered.

**Next.js 16 compatibility (README, index.ts line 614)**
Next.js 16 renamed `middleware.ts` to `proxy.ts` and deprecated `export const middleware` in favor of `export const proxy`. It also dropped Edge Runtime support for the proxy layer, which now runs on Node.js only. The README's headline feature ("100% Edge Runtime Compatible") and all quickstart examples are incorrect for Next.js 16 users. The devDependency pins to `^15.2.4`, so this was never tested against 16. This needs either an explicit Next.js 15-only scope in the docs, or an update to support 16.

**API key logged on refresh failure (line 481)**
```ts
refreshUrl.searchParams.set("key", options.apiKey); // line 465
console.error(refreshUrl.toString(), refreshResponse.status, ...); // line 481
```
The API key is a query param on `refreshUrl` and gets logged verbatim on any non-OK refresh response. In production with centralized log aggregation this exposes the key to anyone with log read access. Fix: log `refreshUrl.origin + refreshUrl.pathname` instead.

**tenantId documented as function but typed as string (README line 125, index.ts line 506)**
The README documents `tenantId: string | ((req: NextRequest) => string | Promise<string>)` but `Config` only has `tenantId?: string`. `normalizeConfig` passes it through without resolving. A user following the docs gets a TypeScript compile error. Either implement the function form or remove it from the docs.

**Empty error message (line 580)**
```ts
if (!normalizedConfigurations.length) throw new Error("");
```
An empty error string gives callers nothing to debug. Should describe what is missing and how to fix it.

---

**Should fix** *(surfaced by automated security review, flagging for verification)*

**JWKS eviction lock not released on network error (lines 218-225)**
If `getFirebaseAuthJwks(cache, true)` at line 221 throws, the lock in cache appears to be left set for the full 30-second TTL with no release path. There is no `finally` block. If that is correct, during that 30-second window every request that gets `ERR_JWKS_NO_MATCHING_KEY` would take the `else` branch and re-throw, rejecting users with a newly-rotated key until the TTL expires. A `try/finally` that deletes the lock on failure would address this if it is indeed a gap.

**Unhandled TypeError from new URL(finalTargetParam) (line 347)**
Line 344 guards against null/empty `finalTarget`, but a non-URL string causes `new URL()` to throw with no surrounding try/catch at this call site. This would surface as a 500 rather than a 400. Worth verifying whether Next.js catches this upstream.

**defaultMemoryCache and serverless (line 59)**
The default in-memory LRU is per-process and empty on every cold start. Worth a note in the docs clarifying that the JWKS eviction lock and JWT caching only work as intended when a shared external cache (Redis, Vercel KV, etc.) is provided. The README currently frames custom caching as optional for performance, which may undersell when it is actually needed.

---

**Minor**

- **`@firebase/util` internal import (line 13):** `getDefaultAppConfig` is from an internal Firebase package with no public stability guarantees. If there is no public-API equivalent, a comment acknowledging the fragility would help.
- **`exp=0` falsy check (line 200):** `if (jwtPayload.exp && ...)` treats `exp=0` as falsy and skips verification. Should be `jwtPayload.exp !== undefined`. Low real-world impact since Firebase never issues `exp=0`, but the guard is logically wrong.
- **`isEmulatorRefreshToken` uses internal Firebase property (line 157):** Relies on `_AuthEmulatorRefreshToken`, which is undocumented. Worth a comment acknowledging it could break if the emulator changes its refresh token format.
- **TODO committed (line 542):** `// TODO don't override string value` should be resolved or removed before merge.
- **Self-referential comment (line 17):** The comment links to this PR itself; remove before merge.
- **`createToken` test helper duplicated (multiple test files):** Defined with slight divergences across `middleware_verify`, `middleware_cache`, and `middleware_refresh`. A shared test utility file would keep them consistent.

---

**Design questions for the team**

1. Is this intended to be a standalone npm package (`firebase-cookie-middleware`) or a `reactfire/nextjs` export? The PR does not modify the root build config, so the publication path is unclear.
2. Should `jose` be bundled (adds around 50KB) or peer-depended? Given that `jose` has had breaking changes across major versions, a tight peer dep range may be safer than bundling.
3. Does the `tenantId` function form need to be implemented before merge, or is it deferred?

src/nextjs/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "peerDependencies": {
+    "firebase": "*",
+    "next": "*"
+  },
+  "devDependencies": {
+    "@upstash/redis": "^1.38.0",
+    "@vitest/coverage-v8": "^4.1.9",
+    "esbuild": "^0.25.1",
+    "firebase": "^11.5.0",
+    "jose": "^6.0.10",
```
tyler-reitz, line 26: `jose` is imported at runtime throughout `index.ts` but listed here as a `devDependency` — it will not be installed in production. Move to `dependencies`, or add to `peerDependencies` to match the README install instructions (`npm install firebase-cookie-middleware jose lru-cache`).

src/nextjs/index.ts
```diff
@@ -0,0 +1,614 @@
...
+      headers: {
+        "Content-Type": "application/x-www-form-urlencoded",
+      },
+    });
+  } catch (e) {
+    console.error("Refresh token request failed.", e);
+    return logout();
+  }
+  if (!refreshResponse.ok) {
+    console.error(refreshUrl.toString(), refreshResponse.status, refreshResponse.statusText);
```
tyler-reitz, line 481: This logs `refreshUrl.toString()` which includes `?key=<apiKey>` set at line 465. The Firebase API key is exposed in plaintext in server logs on any refresh failure. Log `refreshUrl.origin + refreshUrl.pathname` instead.

src/nextjs/index.ts
```diff
@@ -0,0 +1,614 @@
...
+    if (newIdToken) response.cookies.set({ ...ID_TOKEN_COOKIE, value: newIdToken });
+    if (newRefreshToken) response.cookies.set({ ...REFRESH_TOKEN_COOKIE, value: newRefreshToken });
+    return response;
+  };
+  return [undefined, decorateNextResponse, jwtPayload];
+}
+
+export type Config = {
+  emulator?: boolean | string;
+  tenantId?: string;
```
tyler-reitz, line 506: The README documents `tenantId` as accepting `string | ((req: NextRequest) => string | Promise<string>)` but the type here only allows `string`. A user following the docs gets a TypeScript compile error. Either implement the function form or update the README to match.

src/nextjs/index.ts
```diff
@@ -0,0 +1,614 @@
...
+    );
+    if (defaultAppOptions && !normalizedConfigurations.find((it) => it.appName === "[DEFAULT]")) {
+      normalizedConfigurations.push({
+        appName: "[DEFAULT]",
+        firebaseOptions: defaultAppOptions,
+        emulatorHost: process.env.FIREBASE_AUTH_EMULATOR_HOST,
+        cache: defaultMemoryCache,
+      });
+    }
+    if (!normalizedConfigurations.length) throw new Error("");
```
tyler-reitz, line 580: `new Error("")` gives callers nothing to debug. Should describe what is missing, e.g. `"No Firebase configuration found. Pass a Config to composeMiddleware or initialize a Firebase app first."`

src/nextjs/index.ts
```diff
@@ -0,0 +1,614 @@
...
+    }
+    if (!finalResponse)
+      finalResponse = await Promise.resolve(innie(request, defaultUser, allUsers));
+    return decorators.reduce(
+      (prev, current) => current(prev),
+      finalResponse || NextResponse.next(),
+    );
+  };
+
+export const middleware = composeMiddleware(() => {});
```
tyler-reitz, line 614: Next.js 16 deprecated `export const middleware` in favor of `export const proxy` (and `middleware.ts` → `proxy.ts`). It also dropped Edge Runtime support for the proxy layer, which now runs on Node.js only. This export name will be silently ignored by Next.js 16, meaning no auth middleware runs. Needs either explicit Next.js 15-only scoping in the docs, or an update to support 16.

tyler-reitz:
Closing in favor of #739, which carries forward all work from this PR with security hardening and Next.js 16 compatibility. Thanks to @jamesdaniels for the original implementation.

PR #647, opened 2025-05-08
Updating deps for node 20

christhompsongoogle:


jhuleatt:
The "Check if Reference Docs need to be regenerated" job is ok to fail for now. There have been a bunch of breaking TypeDoc updates that would make the diff in this PR gigantic if we ran it.

Instead, locally I pulled the old version of typedoc, forced npm to install despite peer deps warnings with TS, and committed those updated files.

tyler-reitz:
Thanks for contributing, @christhompsongoogle! Main has moved considerably since this was opened and is now ahead on all of these (Firebase 12 peerDep, TypeScript 6, Vite 8, Vitest 4, CI on Node 22/24). Apologies for not catching this sooner. Closing as superseded.

PR #731, opened 2026-07-08
fix: seed useUser with auth.currentUser to prevent undefined on initial render

### LINKED ISSUE #582: useUser() returns undefined unexpectedly on initial render
### Version info
**React:**
18.2.0
**Next:**
13.2.4
**Firebase:**
9.22.1
**ReactFire:**
4.2.3

### Steps to reproduce
I initialize reactfire using the code below. As you can see I have a guard checking if the user is signed in. And only if the user is signed in the children/pages are rendered. 
```tsx
function AuthWrapper(props: any) {
  const app = useFirebaseApp();
  const auth = getAuth(app);
  return <AuthProvider sdk={auth}>{props.children}</AuthProvider>;
}

function AuthCheck(props: React.PropsWithChildren<{ fallback: JSX.Element }>) {
  const router = useRouter();
  const { status, data: signInCheckResult } = useSigninCheck();

  if (status === "loading") {
    return props.fallback;
  }
  if (signInCheckResult?.signedIn === true) {
    return props.children as JSX.Element;
  } else {
    return router.push("/login");
  }
}

export default function App({
  Component,
  pageProps
}: MyAppProps) {
  return (
    <FirebaseAppProvider firebaseConfig={firebaseConfig}>
      <AuthWrapper>
        <AuthCheck fallback={<PageLoader />}>
          <TeamProvider>
            <Component {...pageProps} />
          </TeamProvider>
        </AuthCheck>
      </AuthWrapper>
    </FirebaseAppProvider>
  );
}
```

In my page component, I try to retrieve the user, but on the initial render, the user is undefined. This results in useIdtokenResult throwing an error (Error: you must provide a user) since no user is given and that is required. I need useIdTokenResult, because I need the values of the custom claims.
```tsx
export default function EditRouteForm() {
  const { data: idTokenResult } = useIdTokenResult(user as User, false, {
    initialData: { claims: { group: "", roles: [] } },
  });
  const { group, roles } = idTokenResult.claims
 /* ...... */
}
```

### Expected behavior
Because of my auth guard in the AuthCheck component, I'm expecting useUser to always return a user.

### Actual behavior
The initial render of useUser always returns undefined.

tyler-reitz:
## Problem

Since 4.2.3 (#539 removed the \`initialData\` seeding), \`useUser()\` returns \`{status: 'loading', data: undefined}\` on the first render when called in a child component whose parent only called \`useSigninCheck()\` as a guard.

The root cause: \`useSigninCheck\` and \`useUser\` use different observable IDs (\`auth:signInCheck:...\` vs \`auth:user:...\`). When the guard resolves and renders the child, the child's \`useUser()\` creates a fresh \`SuspenseSubject\` that hasn't received a value yet. Firebase's \`onAuthStateChanged\` fires asynchronously, so the first snapshot is always \`loading\`.

This breaks patterns like:

\`\`\`tsx
function AuthGuard({ children, fallback }) {
  const { status, data: signInCheckResult } = useSigninCheck();
  if (status === 'loading') return fallback;
  if (signInCheckResult.signedIn) return children;
  return <Redirect to="/login" />;
}

function ProfilePage() {
  const { data: user } = useUser(); // undefined on first render
  const { data: idTokenResult } = useIdTokenResult(user as User, false); // throws
}
\`\`\`

## Fix

Restore the synchronous seed: if \`auth.currentUser\` is truthy (a signed-in user is available), set it as \`initialData\` before passing to \`useObservable\`. This means \`useUser()\` returns the user immediately on the first render without waiting for the async observable.

The check is \`if (auth.currentUser)\` (truthy, not \`!== undefined\`) to avoid the ambiguous case where \`currentUser\` is \`null\` before auth has finished loading from storage. If we seeded \`null\` in that case, callers would incorrectly see a signed-out state during initialization.

#539's objection was that \`currentUser\` was undocumented tri-state. That ground has shifted: the SDK now explicitly documents \`currentUser: User | null\` and exposes \`authStateReady()\`. Truthiness is still the right check here since \`authStateReady()\` is async and can't help a synchronous first render.

## Testing

Added a regression test: \`synchronously returns the current user without waiting for the observable\`. The test installs \`NEVER\` as the \`auth:user\` observable before rendering, so any synchronous user value on first render must come from the \`initialData\` seeding path, not from the observable. Without the fix: \`status: 'loading'\`. With the fix: \`status: 'success'\`, \`data: auth.currentUser\`.

Fixes #582

### DIFF
--- a/src/auth.tsx
+++ b/src/auth.tsx
@@ -31,7 +31,7 @@ export function useUser<T = unknown>(options?: ReactFireOptions<T>): ObservableS
   // synchronously on the first render without waiting for the async observable.
   // We only do this when currentUser is truthy to avoid masking the uninitialized
   // (null before auth has loaded from storage) case as "signed out".
-  if (auth.currentUser) {
+  if (auth.currentUser && !('initialData' in _options) && !('startWithValue' in _options)) {
     _options.initialData = auth.currentUser as unknown as T;
   }
 
--- a/src/auth.tsx
+++ b/src/auth.tsx
@@ -25,7 +25,17 @@ export function useUser<T = unknown>(options?: ReactFireOptions<T>): ObservableS
   const observableId = `auth:user:${auth.name}`;
   const observable$ = user(auth);
 
-  return useObservable(observableId, observable$, options);
+  const _options: ReactFireOptions<T> = { ...options };
+
+  // If a user is already signed in, seed initialData so consumers see the user
+  // synchronously on the first render without waiting for the async observable.
+  // We only do this when currentUser is truthy to avoid masking the uninitialized
+  // (null before auth has loaded from storage) case as "signed out".
+  if (auth.currentUser) {
+    _options.initialData = auth.currentUser as unknown as T;
+  }
+
+  return useObservable(observableId, observable$, _options);
 }
 
 export function useIdTokenResult(user: User, forceRefresh = false, options?: ReactFireOptions<IdTokenResult>): ObservableStatus<IdTokenResult> {
--- a/docs/reference/classes/ReactFireError.md
+++ b/docs/reference/classes/ReactFireError.md
@@ -6,7 +6,7 @@
 
 # Class: ReactFireError
 
-Defined in: [src/index.ts:12](https://github.com/FirebaseExtended/reactfire/blob/main/src/index.ts#L12)
+Defined in: [src/index.ts:12](https://github.com/tyler-reitz/reactfire/blob/main/src/index.ts#L12)
 
 ## Extends
 
@@ -18,7 +18,7 @@ Defined in: [src/index.ts:12](https://github.com/FirebaseExtended/reactfire/blob
 
 > **new ReactFireError**(`code`, `message`, `customData?`): `ReactFireError`
 
-Defined in: [src/index.ts:15](https://github.com/FirebaseExtended/reactfire/blob/main/src/index.ts#L15)
+Defined in: [src/index.ts:15](https://github.com/tyler-reitz/reactfire/blob/main/src/index.ts#L15)
 
 #### Parameters
 
@@ -60,7 +60,7 @@ Defined in: node\_modules/typescript/lib/lib.es2022.error.d.ts:24
 
 > `readonly` **code**: `string`
 
-Defined in: [src/index.ts:15](https://github.com/FirebaseExtended/reactfire/blob/main/src/index.ts#L15)
+Defined in: [src/index.ts:15](https://github.com/tyler-reitz/reactfire/blob/main/src/index.ts#L15)
 
 ***
 
@@ -68,7 +68,7 @@ Defined in: [src/index.ts:15](https://github.com/FirebaseExtended/reactfire/blob
 
 > `optional` **customData?**: `Record`\<`string`, `unknown`\>
 
-Defined in: [src/index.ts:15](https://github.com/FirebaseExtended/reactfire/blob/main/src/index.ts#L15)
+Defined in: [src/index.ts:15](https://github.com/tyler-reitz/reactfire/blob/main/src/index.ts#L15)
 
 ***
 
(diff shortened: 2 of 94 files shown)


src/auth.tsx
```diff
@@ -25,7 +25,17 @@ export function useUser<T = unknown>(options?: ReactFireOptions<T>): ObservableS
...
   const observable$ = user(auth);
 
-  return useObservable(observableId, observable$, options);
+  const _options: ReactFireOptions<T> = { ...options };
+
+  // If a user is already signed in, seed initialData so consumers see the user
+  // synchronously on the first render without waiting for the async observable.
+  // We only do this when currentUser is truthy to avoid masking the uninitialized
+  // (null before auth has loaded from storage) case as "signed out".
+  if (auth.currentUser && !('initialData' in _options) && !('startWithValue' in _options)) {
```
armando-navarro, line 34: Consider documenting the behavior change where users will see it: one JSDoc sentence on `useUser` (first render is `success` for already-signed-in users; in suspense mode, signed-in users no longer suspend) + a docs regen. The suspense change is arguably the biggest surface change and it's currently only visible by reading this code comment.

src/auth.tsx
```diff
@@ -25,7 +25,17 @@ export function useUser<T = unknown>(options?: ReactFireOptions<T>): ObservableS
...
+    _options.initialData = auth.currentUser as unknown as T;
```
armando-navarro, line 35: `initialData`'s type collapses to `any`, so this assigns with no cast at all (verified with the repo's own tsc). If you keep a cast, the codebase's existing idiom is `as any as`.

armando-navarro (approved):
Two asks, non-blocking since I'm approving:

1. **Add a regression test that fails on main.** The test you cite passes on main in test suite runs, so nothing today stops a future #539-style removal from silently regressing #582 again. The guard-pattern structure works even mid-suite because `useSigninCheck` and `useUser` cache under different ids: parent renders a `useSigninCheck` guard; once `signedIn`, mount a child calling `useUser()`; assert the child's first-render `data` is the user. Happy to share the exact test I used.
2. **Correct the Testing section of the PR body** (it becomes the squash-commit message): "All 14 auth tests pass" is equally true on main, and the cited test only discriminates when run in isolation. With ask 1 done, the body can simply point at the new test.

A couple more non-blocking notes that didn't fit as inline comments:

- Is it Intentional that `useSigninCheck` stays unseeded? Post-merge, the two auth hooks disagree on first-render state for a signed-in user (`useUser` → instant `success`, `useSigninCheck` → `loading` flash). jhuleatt's comment in #582 pointed at the two hooks using different observables as the root problem; warming/sharing from `useSigninCheck` (no-claims path) would close the asymmetry and has no `currentUser`-timing windows at all. Fine as a follow-up — asking mainly whether it's on your radar.
- Since #539's objection was "undocumented tri-state," worth one line in the body noting the ground has shifted: the SDK now documents `currentUser: User | null` and exposes `authStateReady()`. Truthiness is still a fine choice here (`authStateReady()` is async and can't help a synchronous first render), but pre-empting the #539 argument in the body will save a review cycle if Jeff looks at this.
- A component that mounts while `signOut()` is in flight will now show the outgoing user on its first frame and see no `loading` status for that mount (`success(user)` → `success(null)`). Main already showed a stale signed-in frame one paint later in the same window, so this is a one-frame delta, not a new failure mode — but if someone ever reports "flash of signed-in UI during logout," this is where it lives.

Approving.

tyler-reitz:
Added a regression test in 670733a that should address the ask.

The approach: before rendering, we replace the `auth:user` observable in the preloaded cache with `NEVER` (an observable that never emits). This means any synchronous user value on first render must come from the `initialData` seeding path (the fix), not from the observable. Without the fix the first render sees `status:'loading'`; with it, `status:'success'`.

One tradeoff worth noting: the test deletes the cache entry via `(globalThis as any)._reactFirePreloadedObservables` and hardcodes the internal key format `auth:user:${auth.name}`. The global is deliberately on `globalThis` for cross-module sharing, so it's stable in practice, but the cache key is an internal detail. If that format changes, the test would fail on main for the wrong reason rather than silently passing. Happy to extract a test utility (`_deletePreloadedObservable`) from `useObservable.ts` if you'd prefer to avoid the raw global access.

tyler-reitz:
Re: `useSigninCheck` staying unseeded — yes, it's on our radar. The asymmetry is real and the shared-observable approach you described (warm `useSigninCheck`'s no-claims path, let `useUser` read from the same cache entry) would be the cleaner fix. Tracking it as a follow-up so it doesn't get lost.

tyler-reitz:
Code review also surfaced a pre-existing bug that this PR exercises more frequently: #736 (`useObservable` mutates the shared `_immutableStatus` snapshot in-place). Filed separately — not blocking merge.

PR #732, opened 2026-07-08
fix: replace JSX.Element with React types for React 19 compatibility

### LINKED ISSUE #637: Support for React 19 types in AuthProvider
### Version info

**React:** 19

**Firebase:** 11.0.2

**ReactFire:** 4.2.3

**Other (e.g. Node, browser, operating system) (if applicable):** Node 20

### Problem

As of [this change to `react/@types`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/71411), `reactfire` no longer type checks. It errors like:

```
Type error: 'AuthProvider' cannot be used as a JSX component.
  Its type '(props: PropsWithChildren<{ sdk: Auth; }>) => Element' is not a valid JSX element type.
    Type '(props: PropsWithChildren<{ sdk: Auth; }>) => Element' is not assignable to type '(props: any) => ReactNode | Promise<ReactNode>'.
      Type 'Element' is not assignable to type 'ReactNode | Promise<ReactNode>'.
```

### Expected behavior

It type checks against react 19.

tyler-reitz:
## Problem

Since React 19 removed \`JSX.Element\` from the global \`JSX\` namespace, components in reactfire that return \`JSX.Element\` no longer type-check with React 19 consumers:

\`\`\`
Type error: 'AuthProvider' cannot be used as a JSX component.
  Its type '(props: PropsWithChildren<{ sdk: Auth; }>) => Element' is not a valid JSX element type.
    Type 'Element' is not assignable to type 'ReactNode | Promise<ReactNode>'.
\`\`\`

## Fix

Replace all \`JSX.Element\` annotations with explicit \`React.*\` types. Changed in 4 places:

- \`src/sdk.tsx\` — \`SdkProvider\` return type → \`React.ReactElement\`
- \`src/auth.tsx\` — \`AuthCheck\` return type → \`React.ReactElement\`
- \`src/performance.tsx\` — \`SuspenseWithPerf\` return type → \`React.ReactElement\`
- \`src/storage.tsx\` — \`INTERNALStorageImage\` return type and \`placeHolder\` prop → \`React.ReactNode\` (appropriate since a placeholder can be a string, number, null, etc.)

Also adds a CI \`type-check\` matrix job that runs \`tsc\` against both \`@types/react@18\` and \`@types/react@19\` to prevent regressions.

Fixes #637

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -79,6 +79,27 @@ jobs:
           ./reactfire-${{ github.run_id }}/unpack.sh
       - name: Run tests
         run: npm run test
+  type-check:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        react: ["18", "19"]
+      fail-fast: false
+    name: Type check (React ${{ matrix.react }})
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup node
+        uses: actions/setup-node@v4
+        with:
+          node-version: '24'
+          cache: 'npm'
+      - name: Install deps
+        run: npm ci
+      - name: Install React ${{ matrix.react }} types
+        run: npm install --no-save react@${{ matrix.react }} react-dom@${{ matrix.react }} @types/react@${{ matrix.react }} @types/react-dom@${{ matrix.react }}
+      - name: Type check
+        run: npx tsc --noEmit
   publish:
     runs-on: ubuntu-latest
     name: Publish (NPM)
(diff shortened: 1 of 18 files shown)


armando-navarro (approved):
The PR description might be a little misleading now with regard to the storage change, which becomes the squash merge commit body.  Very minor, but might want to update that before merging.

One possible follow-up issue I found is that CI only type-checks against `@types/react 18`, so nothing stops a global `JSX.Element` from creeping back in and re-breaking `React 19` consumers. A small CI step that packs the build and runs `tsc` over a fixture with `@types/react@19` (and 18) would lock this fix in.

Approving the PR.

PR #735, opened 2026-07-10
fix: surface observable errors via status instead of re-throwing

### LINKED ISSUE #535: useStorageDownloadURL crashes when there is no file in firebase storage
### Version info

**React:**
17.0.2
**Firebase:**
9.6.6
**ReactFire:**
4.2.1
nodejs v16.16.0

### Steps to reproduce

in the code 
```
  const storage = useStorage();
  const catRef = ref(storage, 'kyc/upload/wy_id');
  const { status, data: imageURL } = useStorageDownloadURL(catRef);
```

### Expected behavior
if the file is not there, the status should reflect

### Actual behavior

the website is blank and in the browser's console:
Uncaught FirebaseError: Firebase Storage: Object 'kyc/upload/wy_id' does not exist. (storage/object-not-found)

### LINKED ISSUE #540: cannot handle error from useFirestoreCollectionData
### Version info

**React:** 18.2.0

**Firebase:** 9.9.0

**ReactFire:** 4.2.2

**Next.js** 12.2.2

### Test case

use `useFirestoreCollectionData` and handle `FirebaseError: Missing or insufficient permissions.`

### Steps to reproduce

- Set permissions in firebase console and set a rule that doesn't allow the query for a user
- Make the query
```
const { status, data: users, error } = useFirestoreCollectionData(usersQuery);
```

### Expected behavior

The hook doesn't throw and the error is in `error` variable

### Actual behavior
The hook throws and error cannot be handled without try catch (that should't be used?)
https://github.com/FirebaseExtended/reactfire/blob/9754f8690389feaa1a7a65348aab1567857bca72/src/useObservable.ts#L120-L134

### LINKED ISSUE #743: useIdTokenResult fires a redundant token fetch on every re-render
## Problem

`useIdTokenResult` constructs its observable source eagerly:

```ts
const observable$ = from(user.getIdTokenResult(forceRefresh));
```

`from(promise)` is called at hook-render time, so every re-render that doesn't hit the observable cache fires a discarded `getIdTokenResult` network request. This is the same pattern fixed in `useStorageDownloadURL` by #735.

## Fix

Wrap in `defer()` so the promise is constructed at subscription time, not render time:

```ts
const observable$ = defer(() => user.getIdTokenResult(forceRefresh));
```

## Related

- #735 fixed the same pattern in `useStorageDownloadURL`
- `useCallableFunctionResponse` has a more severe variant (see #744)

tyler-reitz:
## Problem

`useObservable` unconditionally re-throws errors from the underlying observable, making `status: 'error'` unreachable in practice. In non-suspense mode (the default), consumers who want to handle errors locally can't:

```ts
const { status, error } = useStorageDownloadURL(ref);
if (status === 'error') return <NotFound />; // never reached, error was already thrown
```

The original code carried a `TODO(jhuleatt)` questioning whether the unconditional throw was correct.

## Fix

Error handling now matches the mode the consumer opted into:

- **Suspense mode** (`suspense: true`): errors still throw, so React Error Boundaries catch them. Unchanged.
- **Non-suspense mode** (default): errors are returned via `status: 'error'` for local handling. If a hook emits and then errors, `status` becomes `'error'` while `data` keeps its last value.

This keeps v5's **flat `ObservableStatus`** (`data: T`, `error: Error | undefined`) from #750; it does not reintroduce the discriminated-union types from the original version of this PR.

## Docs

Adds an **Error Handling** section to `docs/use.md` documenting both modes, and removes the interim upgrade-guide entry.

## Breaking change, targets v5

This changes default error behavior, so per Jeff's review it targets the **`v5`** branch rather than a v4 patch. Consumers relying on an Error Boundary in non-suspense mode will need an explicit `if (status === 'error') throw error;`.

## Related

- The non-breaking `defer()` fix for the eager-request bug was **split into #757** (v4 patch to `main`, closes #743/#744). The storage `defer` hunk still present here dissolves once #757 lands and `v5` forward-integrates `main`.
- No automatic retry once an observable errors (pre-existing behavior), tracked in #742.

Addresses #535 and #540 (these close when v5 ships).


src/useObservable.ts
```diff
@@ -132,9 +131,9 @@ export function useObservable<T = unknown>(observableId: string, source: Observa
     update.hasEmitted = true;
   }
 
-  // throw an error if there is an error
-  // TODO(jhuleatt) this is the current, tested-for, behavior. But do we actually want it?
-  if (update.error) {
+  // In suspense mode, throw errors so React Error Boundaries can catch them.
+  // In non-suspense mode, surface errors via status so consumers can handle them locally.
+  if (suspenseEnabled && update.error) {
```
armando-navarro, line 134: This guard is what makes `status === 'error'` reachable in default mode — and the narrowed `ObservableStatusError.isComplete: true` is `false` at runtime on every path through here. Details under "Type contract" in my review.

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "4.2.3",
+  "version": "4.3.0",
```
armando-navarro: See "Semver" in my review — the description says breaking, this says minor. Worth an explicit 4.3.0-vs-5.0.0 decision (this number only drives `exp` canary versions either way).

test/useObservable.test.tsx
```diff
@@ -184,6 +184,16 @@ describe('useObservable', () => {
       spy.mockRestore();
     });
 
+    it('surfaces errors via status in non-suspense mode', async () => {
```
armando-navarro: This tests non-suspense behavior but sits in the `Suspense Mode` describe block. The other coverage this PR should pin down is listed under "Tests" in my review.

armando-navarro (changes requested):
Verified this end-to-end locally — the fix itself is right.

Requesting changes on three things, all in service of this being the breaking change the description says it is:

## Documentation
- This changes default error behavior for every hook, and right now the migration story exists only in this PR description (which becomes a squash-commit message).
- At minimum: a JSDoc update on `useObservable`/`ObservableStatus` describing the mode split (which flows into `docs/reference` — will need a docs regen)
- An error-handling example in `docs/use.md` (its current storage example checks only `loading`, so after this merges it silently renders `` on error — the docs would be teaching the silent-failure pattern)
- An entry in `docs/upgrade-guide.md`.

## Type contract
- `ObservableStatusError` declares `isComplete: true`, but runtime is `false` in every error case I tested — the `@ts-expect-error` in `SuspenseSubject._updateImmutableStatus` acknowledges it.
- This was latent while errors always threw; now that `status === 'error'` is the recommended branch, TypeScript narrows consumers into trusting a value that's wrong at runtime. Either make it `isComplete: boolean` or set the runtime flag on error.

## Tests for what this PR actually promises
- The single added test uses a synthetic `throwError` with explicit `suspense: false`.
- Missing:
  - the marquee #535 path (`useStorageDownloadURL`, nonexistent object, default mode, emulator);
  - bare default (`useObservable(id, obs$)` with no suspense option — the headline case); the `FirebaseAppProvider suspense={true}` context path still throwing (distinct code branch, no error test);
  - and the late-error contract (emit-then-error leaves `status: 'error'` with stale `data` still readable — real behavior, currently undocumented and untested;
    - if it's intended, a test locks it in; if not, that's a bug to fix). I ran all of these locally and they behave as described — happy to share the exact tests.

## Two decisions you might want stated explicitly in the PR (not code changes):

- **Semver.** The description says breaking; the bump says 4.3.0 (minor). Repo precedent for breaking was a major (3→4, the Firebase v9 change).
  - There's a fair argument this is a spec-violation fix (`status: 'error'` was always documented and typed, just unreachable) — but that call should be made on purpose, not by the version number in the diff.
  - Note the committed bump only affects `exp` canary numbering either way; tags drive real releases.
- **Error recovery.** Once a mounted cache entry errors, nothing un-errors it (the cache reset timer dies on first subscription).
  - That's pre-existing — but this PR directs users to build `status === 'error'` UI around a state with no retry path.
  - Fine to punt to a follow-up issue; it should be a stated decision.

## Non-blocking notes
- `StorageImage` now swallows errors silently (placeholder forever, `error` never read) where it used to crash — worth either reading `error` (log or render slot) or an explicit "known limitation" note;
- the new test sits in the `Suspense Mode` describe block but tests non-suspense (move it, and don't assert `isComplete: true` on error — it's `false`);
- #540's thread proposed an explicit opt-in flag (`throwErrorInObservable`) rather than coupling to suspense mode — your coupling is defensible and simpler, just worth acknowledging the divergence in the thread when this closes it.

docs/upgrade-guide.md
```diff
@@ -1,3 +1,27 @@
...
+**If you rely on a React Error Boundary to catch Firebase errors in non-suspense mode**, you must add an explicit re-throw in your component:
+
+```tsx
+const { status, error } = useStorageDownloadURL(ref);
+if (status === 'error') throw error; // re-throw to reach your Error Boundary
+```
+
+**If you already check `status` before using `data`**, no change is needed.
+
+Note: once an observable errors there is no automatic retry. If you need retry behavior, unmount and remount the component.
```
armando-navarro: This line is incorrect — remounting doesn't retry. Verified directly: remounting the same `observableId` with a healthy replacement source still returns the original error, since the errored subject stays cached under that id. Suggested fix in my review: state that a new `observableId` is today's only workaround, with retry as the planned follow-up.

test/useObservable.test.tsx
```diff
@@ -124,10 +124,46 @@ describe('useObservable', () => {
 
       act(() => observable$.next('val'));
       expect(result.current.isComplete).toEqual(false);
-      
+
       act(() => observable$.complete());
       await waitFor(() => expect(result.current.isComplete).toEqual(true));
     });
+
+    it('surfaces errors via status in non-suspense mode', async () => {
```
armando-navarro, line 134: The #535 marquee scenario (`useStorageDownloadURL` on a nonexistent object, default mode, storage emulator) still has no test pinning it — this is the PR's reason for existing. Nice to have before merge, not blocking.

src/useObservable.ts
```diff
@@ -71,7 +71,7 @@ export interface ObservableStatusSuccess<T> extends ObservableStatusBase<T> {
 
 export interface ObservableStatusError<T> extends ObservableStatusBase<T> {
   status: 'error';
-  isComplete: true;
+  isComplete: boolean;
```
armando-navarro: Minor, not blocking: could be the precise `false` here rather than `boolean` — error-after-complete can't happen. Or just delete the override since it now matches the base type.

armando-navarro (approved):
Thanks for the thorough response - nearly everything landed.  Take this review with a grain of salt, as there are significant gaps in my knowledge here that could lead to incorrect conclusions.

I'm approving this PR with a couple additional suggestions:

## The upgrade guide's retry sentence is wrong
- "If you need retry behavior, unmount and remount the component" is incorrect — remounting doesn't retry.
- The errored subject stays in the `globalThis` cache under its `observableId`, and a remount rejoins that same cached subject.
- Verified directly: remounting the same id with a *healthy* replacement source still returns the original error.
- It also contradicts your own PR body's "errored observables do not re-subscribe."
- Suggested replacement: today the only workaround is a new `observableId`; a retry mechanism is the follow-up issue you're already planning.

## The marquee scenario still has no test
- The #535 path — `useStorageDownloadURL` on a nonexistent object, default mode, storage emulator → `status: 'error'` with the `storage/object-not-found` FirebaseError — is the PR's reason for existing.
- It's verified working, but nothing in the suite pins it.
- One emulator test would lock the headline behavior in.
- Nice-to-have, not blocking: the `FirebaseAppProvider suspense={true}` context-path error test from my earlier list would also be welcome.

## Two minor notes (no action required)
- `isComplete` on the error branch could be the precise `false` rather than `boolean` — error-after-complete can't happen, since completion never reaches the tap once the source errors.
  - Alternatively, just delete the override, since it now matches the base type.
- The subscriber `error:` callback in `useObservable` is unreachable (`catchError` upstream converts errors to completion).
  - That branch could carry a comment or be dropped.
  - The real behavior change in this PR is the render-path condition, which is correct.

tyler-reitz (commented):
Fixed. Updated the note to explain that remounting rejoins the cached errored subject, and that changing `observableId` is the only current workaround. A proper retry mechanism is the planned follow-up.

docs/upgrade-guide.md
tyler-reitz: Fixed. Updated the note to explain that remounting rejoins the cached errored subject, and that changing `observableId` is the only current workaround. A proper retry mechanism is the planned follow-up.

test/useObservable.test.tsx
tyler-reitz: Added. Had to pair it with a `defer(() => getDownloadURL(ref))` fix in `storage.tsx` since `getDownloadURL` was firing an eager network request on every render, which leaked an unhandled rejection when no subscriber caught the result. The fix also eliminates redundant requests on re-renders as a bonus.

tyler-reitz (commented):
Most of this has landed: upgrade guide updated with the correct retry explanation, `isComplete` override removed, the marquee test added (had to fix an eager-request bug in `useStorageDownloadURL` first, wrapped with `defer`, which also eliminates redundant network requests on re-renders as a bonus), and the `FirebaseAppProvider suspense={true}` context-path error test added. Semver call is @jhuleatt's.

docs/upgrade-guide.md
```diff
@@ -1,3 +1,27 @@
...
+**If you rely on a React Error Boundary to catch Firebase errors in non-suspense mode**, you must add an explicit re-throw in your component:
+
+```tsx
+const { status, error } = useStorageDownloadURL(ref);
+if (status === 'error') throw error; // re-throw to reach your Error Boundary
+```
+
+**If you already check `status` before using `data`**, no change is needed.
+
+Note: once an observable errors there is no automatic retry. The errored observable remains in the global cache under its `observableId`, so unmounting and remounting the same component rejoins the same errored state. Today the only workaround is to change the `observableId`. A proper retry mechanism is tracked as a follow-up issue.
```
armando-navarro: The corrected retry explanation is accurate now (verified against the cache behavior). One thing: this says a retry mechanism "is tracked as a follow-up issue," but I couldn't find such an issue open or closed. Could you file it and cite the number, or soften to "planned as a follow-up"?

src/storage.tsx
```diff
@@ -28,7 +29,7 @@ export function useStorageTask<T = unknown>(task: UploadTask, ref: StorageRefere
  */
 export function useStorageDownloadURL<T = string>(ref: StorageReference, options?: ReactFireOptions<T>): ObservableStatus<string | T> {
   const observableId = `storage:downloadUrl:${ref.toString()}`;
-  const observable$ = getDownloadURL(ref);
+  const observable$ = defer(() => getDownloadURL(ref));
```
armando-navarro, line 35: This `defer()` is a real fix, not just scope creep: rxfire's `getDownloadURL` is eager, so every re-render fired a discarded request and a rejecting one produced an unhandled rejection (your new test fails without this wrapper). Worth noting the same eager pattern remains in `useIdTokenResult` and `useCallableFunctionResponse` (the callable side-effects per render) — fine to leave for follow-ups, just worth a line in the PR body so it doesn't read as an oversight.

test/useObservable.test.tsx
```diff
@@ -328,5 +366,23 @@ describe('useObservable', () => {
       // if useObservable doesn't re-emit, the value here will still be "Jeff"
       expect(refreshedComp).toHaveTextContent('James');
     });
+    it('throws an error via FirebaseAppProvider suspense context path', () => {
```
armando-navarro, line 369: Verified this exercises the context branch (hook called with no config, so the provider's `suspense={true}` decides). Minor: the sibling test around line 172 adds a `window.addEventListener('error', e => e.preventDefault())` guard to suppress the jsdom uncaught-error noise this kind of throw prints; copying it here would quiet the run. Not blocking.

armando-navarro (approved):
Thanks Tyler, this round landed cleanly. I verified everything from the previous rounds at head 0fde812: the corrected retry note now matches the actual cache behavior, the `isComplete` override removal is a clean no-op (tsc passes), and both new tests are sound. The storage test genuinely pins the #535 scenario against the emulator (3/3 passing runs on my machine, no unhandled rejections), and the context-path test exercises the provider branch rather than per-hook config (18/18 in that file).

The `defer()` change deserves its own callout: I initially flagged it as unrequested scope, but it checks out as a real fix. rxfire's `getDownloadURL` constructs its promise eagerly, so before this change every re-render fired a discarded request, and a rejecting one (missing file) produced an unhandled promise rejection. I confirmed your new test fails without the wrapper, and that the deferred factory executes exactly once across repeated renders.

Two small things to resolve, neither touching code:

### The "tracked as a follow-up issue" sentence

- The upgrade guide now says a retry mechanism "is tracked as a follow-up issue," but I couldn't find such an issue (searched open and closed).
- Could you file it and cite the number, or soften the sentence to "planned as a follow-up"? Otherwise, I think it will be pointing at an issue that doesn't exist.

### The defer fix is partial, which is fine, but worth saying

- The same eager pattern remains in `useIdTokenResult` (`from(user.getIdTokenResult(...))`, a token fetch per render) and `useCallableFunctionResponse` (rxfire's `httpsCallable`, which invokes the cloud function on every render and discards the result).
- Fixing only storage here is reasonable scope control. A sentence in the PR body noting the siblings as known follow-ups would keep it from reading as an oversight, and the callable one may deserve its own issue given it side-effects.

Minor notes, no action needed: the new context test could copy the `window.addEventListener('error', ...)` guard its sibling uses (the run currently prints jsdom uncaught-error noise); the storage test's `(result.current.error as any)?.code` could type the error as `FirebaseError` instead of casting; and the `e021670` commit message describes a removal its diff doesn't perform, which is harmless under squash merge but worth knowing if these commits are ever preserved.

If any of this reads wrong, let me know. Approving.

docs/upgrade-guide.md
```diff
@@ -1,3 +1,27 @@
...
+```tsx
+const { status, error } = useStorageDownloadURL(ref);
+if (status === 'error') throw error; // re-throw to reach your Error Boundary
+```
+
+**If you already check `status` before using `data`**, no change is needed.
+
+Note: once an observable errors there is no automatic retry. The errored observable remains in the global cache under its `observableId`, so unmounting and remounting the same component rejoins the same errored state. Today the only workaround is to change the `observableId`. A proper retry mechanism is tracked in [#742](https://github.com/FirebaseExtended/reactfire/issues/742).
+
+---
```
jhuleatt: Instead of adding this here, can you please add a section to use.md with title `Error Handling` and document the two different ways? (suspense mode vs non-suspense mode)

src/useObservable.ts
```diff
@@ -83,6 +82,17 @@ export interface ObservableStatusLoading<T> extends ObservableStatusBase<T> {
 
 export type ObservableStatus<T> = ObservableStatusLoading<T> | ObservableStatusError<T> | ObservableStatusSuccess<T>;
 
+/**
```
jhuleatt: Thanks for adding this!

src/storage.tsx
jhuleatt, line 35: Is this fix related to the error behavior change? If not, could you please break it out into its own PR so we can include it in a v4 patch release?

jhuleatt (changes requested):
While I we could argue this is **technically** a bugfix, in reality we might break a lot of people that currently use error boundaries if we ship this as a patch. Let's call it a breaking change and save it for v5

tyler-reitz:
@jhuleatt heads-up on process: since you flagged this as breaking and bound for v5, I cut a **`v5`** branch off `main` on upstream to house it rather than let it block, and retargeted this PR there. The intended flow: v4 patches and minors keep landing on `main`, breaking work targets `v5`, `main` forward-integrates into `v5` periodically, and `v5` merges to `main` at the 5.0.0 release. I went with `v5` over `next` to match how we already refer to it.

The non-breaking half of this PR (the `defer()` eager-request fix) is split out to #757 as a v4 patch. Flag anything you'd want done differently, especially around release and canary publishing from `v5`, since that's your call.

armando-navarro (approved):
Thanks for reworking this onto v5, Tyler, and for the Error Handling docs section. I re-verified the error-handling change on the v5 base and it's correct: in non-suspense mode an error now comes back as `status: 'error'` with `data` holding its last value, and in suspense mode it still throws, so Error Boundary users are unaffected.

### What I verified

- The one-line change (`if (update.error)` to `if (suspenseEnabled && update.error)`) is load-bearing: reverting just it fails exactly the three non-suspense error tests while the suspense-context test still passes, so the tests genuinely pin the behavior.
- Suspense mode is unchanged from base: with `suspenseEnabled === true` the guard is equivalent to the old unconditional throw, and `SuspenseSubject` is untouched, so the change affects only the default mode. Traced the throw-a-promise, initialData, and main paths and none swallows an error.
- The "data keeps its last value on error" claim holds through `immutableStatus` (the error handler sets `error` without clearing the last value, and status is computed error-first), and the docs section matches the code.
- `tsc` is clean on `src` and `src + test`, the regenerated reference docs match a fresh `npm run docs`, the `useObservable` suite is 18/18, and the storage `object-not-found` test surfaces the error via `status` end-to-end against the emulator.

### Two cleanups before merge

- The storage `defer` fix (plus its `object-not-found` test and the regenerated storage docs) is still in this PR, but it's also in #757, which you split out at Jeff's request and which I've approved. Against v5 that's a duplication and a likely merge conflict on the same `storage.tsx` line. I'd let #757 land and reach v5, then rebase this on v5 and drop the storage hunk here while keeping the error test. Right now the test depends on the in-PR defer, so it can't just be removed yet.
- The `.gitignore` gained a `CLAUDE.local*` line that isn't related to error handling; probably better in a global gitignore than in the repo.

Optional: the throw inside the `initialData` block is now unreachable on an error (the observable reports `hasValue` once errored, so that block is skipped), so you could drop it to avoid two confusing throw sites, or leave it as defensive.

Approving on the error-handling change.

Issue #749, opened 2026-07-21
Add a published .d.ts diff to the release process to catch breaking type changes


tyler-reitz:
## Problem

4.2.4 shipped as a patch but contained breaking TypeScript type changes with no changelog note. The main one: `ObservableStatus<T>` was refactored from a flat interface (`data: T`) into a discriminated union (`data: T | undefined` unless narrowed on `status`), which breaks the standard destructure-and-use pattern across every data hook, including the documented suspense pattern. It is type-only (no runtime impact), but it reds strict-TS consumer CI on upgrade.

It slipped through because:

- The change came in via #583 ("use `useSyncExternalStore` to sync data"), whose title looked like an internals change, not a public API break.
- It then sat unreleased for ~3 years (v4.2.3 was 2022-08, #583 merged 2023-07).
- 4.2.4 batched 35 PRs of accumulated `main` into one bump, with no step auditing the cumulative public type surface.

## Proposal

Add a release-time (or CI) check that diffs the candidate's emitted types against the last published version:

1. `npm pack` the latest published version, extract `dist/*.d.ts`.
2. `npm pack` the release candidate, extract `dist/*.d.ts`.
3. Diff them. Any non-additive change (removed/narrowed/changed signature) fails the check or requires an explicit "breaking" acknowledgment and a minor/major bump.

This exact diff would have flagged both the `ObservableStatus` union change and the `useFirestoreDocData` widening (#733) immediately.

## Related

- Remediation for the live 4.2.4 release is tracked separately (deprecate + re-cut as 4.3.0 with a migration note).
- Surfaced while reviewing #740.

tyler-reitz:
### End-to-end proof of the type break

Built a fresh consumer project, installed `reactfire` straight from npm, and ran `tsc --strict` on the documented suspense-mode pattern. Only the reactfire version changes between runs; `tsconfig.json` and `app.tsx` are identical.

`app.tsx`:
```tsx
import { useFirestoreCollectionData } from 'reactfire';
import type { Query } from 'firebase/firestore';

interface Post { title: string; }

export function PostList({ postsQuery }: { postsQuery: Query<Post> }) {
  const { data: posts } = useFirestoreCollectionData(postsQuery);
  return posts.map((p) => p.title);
}
```

| reactfire | Code | `tsc --strict` |
|---|---|---|
| 4.2.3 | `posts.map(...)` | green (exit 0) |
[...]

PR #583, opened 2023-06-29
use `useSyncExternalStore` to sync data in `useObservable`

jhuleatt:
### Description

Migrating the core of ReactFire to [`useSyncExternalStore`](https://react.dev/reference/react/useSyncExternalStore), which is the recommended way to sync state between React and something external (in our case RxJS).

The main changes are in `src/useObservable.ts` and `src/SuspenseSubject.ts`.

Also adds a test to `test/firestore.test.tsx` to address #327.

PR #750, opened 2026-07-22
fix: revert ObservableStatus to a strict (non-union) type

tyler-reitz:
## Summary

4.2.4 shipped `ObservableStatus<T>` as a discriminated union (#583), so `data` became `T | undefined` unless narrowed on `status`. That broke the documented pattern of reading `.data` directly (e.g. `const { data } = useFirestoreCollectionData(q); data.map(...)`) across every data hook, including the suspense pattern, and it went out as a patch.

This reverts `ObservableStatus` to its previous strict shape (`data: T`) so the fix ships as a patch, and adds a type-check over the test suite in CI so a regression like this fails at compile time going forward.

## Commits

1. **`test: type-check the test suite in CI`** adds `tsconfig.test.json` (type-checks `src` + `test`) and runs it in the type-check workflow. With the union still in place this turns the check **red** with 28 `TS18048` "possibly undefined" errors on the existing test assertions. Also fixes pre-existing test-only type issues the check surfaces (unused imports, node globals, a `children` prop annotation).
2. **`fix: revert ObservableStatus to a strict (non-union) type`** reverts the type to `data: T`. The check goes **green** (0 errors).
3. **`docs: regenerate reference docs`** for the `ObservableStatus` type change (drops the per-status sub-interface pages, retargets links from `type-aliases/` to `interfaces/`).
4. **`test: strengthen the type guard`** with a typed `useDatabaseObjectData` read (which fails to compile if `ObservableStatus` regresses to a union) and a concretely-typed `useFirestoreDocData` read, per review.

The existing tests already read `.data` off hook results the way consumers do, so once type-checked they catch this class of break with no separate type-regression file.

## What stays correctly typed per hook

The genuinely-optional cases keep their `undefined` / `null` at the hook level, which is where it belongs:

- `useFirestoreDocData` / `useFirestoreDocDataOnce` return `ObservableStatus<T | undefined>` (#733)
- `useUser` returns `ObservableStatus<User | null>`
- `useDatabaseListData` returns `ObservableStatus<T[] | null>`

Reverting the union cleared all 28 possibly-undefined errors, confirming no other hook needed per-hook `undefined`.

## Removed types (for release notes)

Reverting removes the three interfaces 4.2.4 introduced: `ObservableStatusSuccess`, `ObservableStatusError`, `ObservableStatusLoading`. Anyone who adopted them in 4.2.4 should read `data` off `ObservableStatus<T>` directly again. Worth a changelog line on the 4.2.5 release.

## Verification

- Red then green: 28 `TS18048` at commit 1, 0 at commit 2.
- `tsc --noEmit` green on both `src` and `src` + `test`.
- Non-emulator tests (`useObservable`, `firebaseApp`) pass (21/21); emulator suites green in CI.
- Consumer check: packed this branch and type-checked the documented `useFirestoreCollectionData(...).data.map(...)` pattern under `strict`, green here, red on 4.2.4.

Intended to ship as a patch (4.2.5). Tracks #749.


armando-navarro (approved):
Nice fix, along with the CI type-check. I verified the whole thing on current `main`: the revert is faithful to the pre-4.2.4 shape, the red-then-green holds, and types, tests, and docs are all green.

### What I verified

- The revert restores `ObservableStatus<T>` to the flat `data: T` shape (identical to 4.2.3), and the genuinely-optional cases stay where they belong: `useFirestoreDocData`/`DocDataOnce` → `T | undefined`, `useUser` → `User | null`, `useDatabaseListData` → `T[] | null`.
- Red then green: `tsc -p tsconfig.test.json` reports exactly 28 `TS18048` at the guard commit (union still in place) and 0 after the revert. The 28 are real `.data` reads, and the test fixes added no `as any` / `@ts-ignore` / `!` suppressions, so the guard has real teeth.
- `tsc` is clean on `src` and `src + test`; the non-emulator suites are 21/21; the firestore/database/auth emulator suites are green; and the regenerated reference docs match a clean `npm run docs`, so the docs check will pass.

### The one decision that isn't mine: semver

Reverting removes three types that 4.2.4 published (`ObservableStatusSuccess`, `ObservableStatusError`, `ObservableStatusLoading`) and re-narrows `data`. It restores the 4.2.3 contract, so shipping it as a patch to undo a bad patch reads as reasonable to me, but removing exported types is technically breaking, so I would rather Jeff make that call explicitly.

Either way, a changelog line noting those three types are gone would help anyone who adopted them in 4.2.4.

### Smaller notes

- The guard is genuine but not exhaustive: the `useFirestoreDocData` tests use `<any>`, so `.data` is untyped there, and `useDatabaseObjectData`/`useDatabaseListData` are not exercised at all, so a future union reintroduction on those would not be caught. One concretely-typed `.data` read for doc data plus an object/list test would close the gaps. Non-blocking.
- The description says two commits, but there are three (the docs regen is the third, and worth keeping).

Everything code-level checks out, so approving on that, with the semver call left to Jeff.

src/useObservable.ts
```diff
@@ -64,25 +64,6 @@ interface ObservableStatusBase<T> {
...
-  status: 'error';
-  isComplete: true;
-  error: Error;
-}
-
-export interface ObservableStatusLoading<T> extends ObservableStatusBase<T> {
-  status: 'loading';
-  data: undefined;
-  hasEmitted: false;
-}
```
jhuleatt, line 82: This state is valid, isn't it? In that case, are 4.2.4's types a legitimate fix?

jhuleatt: oh, or is TS not smart enough to figure out that this isn't a condition that can be hit in Suspense mode? Is there another way to indicate to TS that Suspense mode won't hit this state?

jhuleatt: Hm, maybe getting fancier with the types for Suspense mode is out of scope for this fix.

jhuleatt (approved):
Thanks!

tyler-reitz:
Thanks Armando! Got both in a new commit: added a useDatabaseObjectData test with a typed .data read (fails to compile if ObservableStatus goes back to a union, confirmed) plus a concretely-typed docData read instead of the <any> ones, no suppressions. Also fixed the commit count and added a "Removed types" note for ObservableStatusSuccess/Error/Loading. Landing as a patch for 4.2.5.

src/useObservable.ts
tyler-reitz: Yeah, it's a valid state and 4.2.4's union modeled it fine, the issue was really that it broke reading .data directly (especially in Suspense where it's always there) and shipped as a patch. Proper Suspense-mode types where data: T is non-optional would be the real fix, but agreed that's out of scope here, so I captured it for V5.

tyler-reitz:
## Final due-diligence: published `.d.ts` surface diff

Packed this branch and diffed the emitted type declarations against both the broken release (4.2.4) and the last-good one (4.2.3), the same release-time check tracked in #749.

**vs 4.2.4 — surgical.** The only `.d.ts` that changes is `useObservable.d.ts` (the ObservableStatus revert + removal of the three sub-interfaces). Every other module, firestore/database/auth/storage/etc., is byte-identical to 4.2.4. No collateral changes.

**vs 4.2.3 — contract restored.** The `ObservableStatus` interface is identical to 4.2.3. The only difference in the whole file is one additive line:
```
- preloadObservable<T>(source, id): SuspenseSubject<T>
+ preloadObservable<T>(source, id, suspenseEnabled?): SuspenseSubject<T>
```
That optional `suspenseEnabled?` comes from #583's useSyncExternalStore runtime (kept intentionally, we reverted only the type). Optional param = additive/non-breaking.

Net: 4.2.3's `ObservableStatus` contract is restored, #733's per-hook `undefined` is preserved, and the only delta from 4.2.3 is a benign additive param.

PR #757, opened 2026-07-23
fix: defer eager requests in storage, auth, and functions hooks

### LINKED ISSUE #743: useIdTokenResult fires a redundant token fetch on every re-render

### LINKED ISSUE #744: useCallableFunctionResponse invokes the cloud function on every re-render
## Problem

`useCallableFunctionResponse` constructs its observable source eagerly via rxfire's `httpsCallable`, which invokes the cloud function at hook-render time. Every re-render that doesn't hit the observable cache fires a real cloud function call that is then discarded.

Unlike `useIdTokenResult` (which wastes a network read), this has a write-side effect: the function executes and bills against quota on every render cycle.

## Fix

Wrap in `defer()` so the function is invoked at subscription time, not render time:

```ts
const observable$ = defer(() => httpsCallable(functionRef, data, options));
```

## Related

- #735 fixed the same pattern in `useStorageDownloadURL`
- #743 tracks the same issue in `useIdTokenResult`

tyler-reitz:
## What

`useStorageDownloadURL`, `useIdTokenResult`, and `useCallableFunctionResponse` each constructed their request in the render body, so the underlying call ran on every render, firing a discarded request each time (and, for a missing storage object, an unhandled promise rejection). This wraps each in rxjs `defer` so the request is created lazily on subscription. Since `useObservable` subscribes once per `observableId`, the request now runs once instead of once per render.

- `useStorageDownloadURL`: `defer(() => getDownloadURL(ref))`
- `useIdTokenResult`: `defer(() => from(user.getIdTokenResult(forceRefresh)))`
- `useCallableFunctionResponse`: `defer(() => obsFactory(options?.data))`

Closes #743, closes #744.

## Why a separate PR

Split out of #735 at Jeff's request: the `defer` fix is a non-breaking bugfix that can ship in a v4 patch, independent of #735's (breaking) error-surface change, which is now slated for v5.

## Testing

Adds a regression test for `useIdTokenResult` using an injected user whose `getIdTokenResult` is a spy: it is called once across re-renders with the wrapper, and 4 times without it (verified locally). The three hooks share the same one-line `defer` pattern through `useObservable`, so this pins the mechanism. Happy to add module-mock call-count tests for the storage and functions hooks too if reviewers would like per-hook coverage.

### DIFF
--- a/docs/reference/functions/AuthCheck.md
+++ b/docs/reference/functions/AuthCheck.md
@@ -8,7 +8,7 @@
 
 > **AuthCheck**(`__namedParameters`): `ReactElement`
 
-Defined in: [src/auth.tsx:257](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L257)
+Defined in: [src/auth.tsx:259](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L259)
 
 ## Parameters
 
--- a/docs/reference/functions/ClaimsCheck.md
+++ b/docs/reference/functions/ClaimsCheck.md
@@ -8,7 +8,7 @@
 
 > **ClaimsCheck**(`__namedParameters`): `Element`
 
-Defined in: [src/auth.tsx:213](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L213)
+Defined in: [src/auth.tsx:215](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L215)
 
 ## Parameters
 
--- a/docs/reference/functions/StorageImage.md
+++ b/docs/reference/functions/StorageImage.md
@@ -8,7 +8,7 @@
 
 > **StorageImage**(`props`): `Element`
 
-Defined in: [src/storage.tsx:78](https://github.com/FirebaseExtended/reactfire/blob/main/src/storage.tsx#L78)
+Defined in: [src/storage.tsx:82](https://github.com/FirebaseExtended/reactfire/blob/main/src/storage.tsx#L82)
 
 ## Parameters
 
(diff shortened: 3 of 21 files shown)


armando-navarro (approved):
Nice, clean fix. I verified all three hooks on current `main` (post-4.2.5): each now creates its request lazily on subscription instead of in the render body, and I confirmed the change is behavior-preserving apart from removing the per-render repeat.

### What I verified

- The per-render waste is real and the fix removes it: with the new test, `getIdTokenResult` is called once across a render plus two re-renders; reverting just the `defer` wrapper makes it fail with four calls. So the test genuinely pins the fix.
- `defer` is safe under this architecture: `SuspenseSubject` subscribes eagerly in its constructor and the source ends in `shareReplay(1)`, so the request still fires at the same moment as before and runs exactly once. Cache identity, the suspense `firstEmission` path, and error surfacing are all unchanged.
- The storage unhandled-rejection fix is real rather than relocated: previously each non-first render created a fresh rejecting promise with no subscriber; now the promise is only created for the single warmup subscription, whose rejection is handled by the existing `tap` + `catchError`. No new unhandled rejection.
- `tsc` is clean on `src` and on `src + test`, and the regenerated reference docs match a fresh `npm run docs` (just the `Defined in` line shifts from the added import).

### One optional follow-up

Only `useIdTokenResult` has a regression test; `useStorageDownloadURL` and `useCallableFunctionResponse` got the identical fix but lean on that test as a proxy.

The mechanism is the same across all three, so this isn't a correctness concern, but a call-count test for the functions hook (where the eager re-call actually invokes a cloud function) would guard against a future regression.

You already offered these in the description, so this is just a +1. Non-blocking.

Approving.

PR #758, opened 2026-07-24
chore: forward-integrate main into v5

tyler-reitz:
Forward-integrates main into v5 (fast-forward, no conflicts) to bring in #757 (defer eager requests in storage/auth/functions hooks).

This lets #735's temporary storage `defer` hunk dissolve on its next rebase, since that hunk was only kept to unblock the storage error test until #757 landed.

No new changes beyond what's already on main: docs regen, src defer wraps in auth.tsx/functions.tsx/storage.tsx, and their tests. tsc (src + test configs) and the functions/auth emulator suites pass locally on Node 22.

### DIFF
--- a/docs/reference/functions/AuthCheck.md
+++ b/docs/reference/functions/AuthCheck.md
@@ -8,7 +8,7 @@
 
 > **AuthCheck**(`__namedParameters`): `ReactElement`
 
-Defined in: [src/auth.tsx:257](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L257)
+Defined in: [src/auth.tsx:259](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L259)
 
 ## Parameters
 
--- a/docs/reference/functions/ClaimsCheck.md
+++ b/docs/reference/functions/ClaimsCheck.md
@@ -8,7 +8,7 @@
 
 > **ClaimsCheck**(`__namedParameters`): `Element`
 
-Defined in: [src/auth.tsx:213](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L213)
+Defined in: [src/auth.tsx:215](https://github.com/FirebaseExtended/reactfire/blob/main/src/auth.tsx#L215)
 
 ## Parameters
 
--- a/docs/reference/functions/StorageImage.md
+++ b/docs/reference/functions/StorageImage.md
@@ -8,7 +8,7 @@
 
 > **StorageImage**(`props`): `Element`
 
-Defined in: [src/storage.tsx:78](https://github.com/FirebaseExtended/reactfire/blob/main/src/storage.tsx#L78)
+Defined in: [src/storage.tsx:82](https://github.com/FirebaseExtended/reactfire/blob/main/src/storage.tsx#L82)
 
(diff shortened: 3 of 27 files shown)


armando-navarro (approved):
Thanks Tyler. I verified this is a clean forward-integration: the single commit here is the squash-merge of #757 (its parent is the current v5 tip), so merging fast-forwards v5 by exactly that one already-reviewed, already-shipped commit, with no conflicts and nothing extra.

I also re-checked the `functions.test.tsx` you added after my #757 approval (it addresses the functions-hook coverage I'd asked about): it passes, and reverting the `defer` makes it fail with four calls, so it's a real regression tripwire.

Two non-blocking notes:

- `main` has moved two commits past this branch since it was cut, so this brings v5 up to #757 but not to main's current tip.
	- In particular it leaves out #760 (now tagged v4.2.6) and the version bump. Since v5 is still a strict ancestor of main, you could re-point this PR at main's current tip and fast-forward the whole way, folding #760 and the bump into one integration instead of needing a second sync right after.
	- If stopping at #757 is deliberate, a note in the description would make that clear.
- The only failing check here is Publish (NPM); every functional check (build, both Node versions, both React type-checks) is green. Could you confirm that's the known publish flakiness rather than a new break?

Approving the integration.

tyler-reitz:
Plan for this integration: going with Armando's suggestion to re-point at main's current tip rather than stopping at #757.

Since v5 is still a strict ancestor of main, this stays a clean fast-forward (no conflicts). Re-pointing now folds in everything that has landed since this branch was cut: #757, #760, the 4.2.6 version bump, and the four Dependabot lockfile bumps (#761-764).

Holding the re-point until Monday so we also capture anything that lands over the weekend and avoid a second sync right after. Re-pointing will move the head off the commit Armando approved, so it will need a fresh review (should be quick given it is a clean fast-forward).

tyler-reitz:
@armando-navarro heads up: I re-pointed and merged this while you're out, so the head moved off the commit you approved. Flagging exactly what changed in case you want to look when you're back.

Your approval was on `38a6a4e` (v5 + #757 only). I fast-forwarded the branch to main's tip `a2090a5` before merging, which added six commits:

- `1892f09` fix: externalize useSyncExternalStore shim to fix App Router crash (#760)
- `7f93210` 4.2.6
- `df95048` / `279d522` / `d33d291` / `a2090a5` Dependabot lockfile-only bumps (#764, #761, #763, #762)

Rationale for folding them in rather than stopping at #757: `v5` was a strict ancestor of `main`, so this stayed a clean fast-forward and picked up 4.2.6 in one sync instead of leaving a second forward-integration to do next week. All six had already been reviewed and merged on `main` on their own.

CI was green on the new head (build, tests on Node 22 and 24, type check on React 18 and 19, CodeQL, CLA). The one red on the commit is `Publish (NPM)`, which is inherited from main's own push run, not from this PR, and is the known-broken publish job Jeff has been working around with manual publishes.

Merged as a merge commit (`aec5b83`) deliberately, not squash or rebase, so `v5` keeps `main` as a real ancestor and future forward-integrations stay conflict-free.

No action needed from you. If anything here looks off, say so and I'll follow up. This also unblocks #735: the temporary storage `defer` hunk should dissolve on the next rebase now that #757 is on `v5`.

PR #760, opened 2026-07-24
fix: externalize useSyncExternalStore shim to fix App Router crash

### LINKED ISSUE #759: App Router crash: reactfire 4.2.4/4.2.5 throw "dynamic usage of require is not supported" on the client
### Summary

Reactfire **4.2.4** and **4.2.5** (current `latest`) crash on the client in Next.js App Router apps that render a data hook (e.g. `useFirestoreDocData`). `use-sync-external-store/shim` (a CJS module) is bundled into our ESM `dist`, which makes the bundler emit a dynamic `require()`. In a strict-ESM browser context it throws at module evaluation of the reactfire chunk:

```
Error: dynamic usage of require is not supported
```

Any App Router page that loads a reactfire data-hook chunk breaks on the client. Server prerender is unaffected (Node has `require`).

### Affected versions

- **4.2.4, 4.2.5**: affected (ship the bundled shim with the dynamic `require`).
- **4.2.3**: clean (predates the current build toolchain; ESM dist imports `react` directly, no `require`).

### Reproduction

1. Next.js App Router app on reactfire 4.2.5.
2. Render a client component using a data hook, e.g. `useFirestoreDocData`.
3. Load the page: the browser console throws `Error: dynamic usage of require is not supported` at module evaluation.

### Root cause

The build-tooling migration (around 4.2.4) began bundling the CJS `use-sync-external-store/shim` into the ESM output, producing a runtime `require()` that does not exist in strict ESM.

### Fix

Externalize `use-sync-external-store/shim` so the consumer's bundler resolves it (it is already a runtime dependency). Non-breaking; the shim's React 16/17 fallback is preserved. Fix PR incoming.

tyler-reitz:
Fixes #759

### Summary

Reactfire **4.2.4 and 4.2.5** crash on the client whenever a data hook renders, in any browser bundle that does not expose `require`. `use-sync-external-store/shim` (CJS) gets bundled into our ESM `dist`, which makes the bundler emit a dynamic `require()`. In a strict-ESM browser context that throws at module evaluation:

```
Calling `require` for "react" in an environment that doesn't expose the `require` function
```

4.2.3 predates the current build toolchain and is clean.

### Fix

Add `use-sync-external-store/shim` to the Vite `externals` so the consumer's bundler resolves it (it is already a runtime `dependency`). This removes the `require` from our dist while keeping the shim's React 16/17 fallback, so it stays non-breaking (no peer-range change).

### Verification

- Built ESM `dist/index.js`: `require` occurrences 3 to 0; clean `import ... from "use-sync-external-store/shim"`.
- Reproduced the live crash with published 4.2.5 in two setups: a Next 16 App Router app (turbopack) and a Vite SPA (both crash client-side at module evaluation).
- Rebuilt both against this fix: Next client chunk no longer carries the throw path; the Vite app renders successfully (hook reaches `status: 'success'`).

Patch-level, non-breaking. Fixes the regression in `latest`. Version bump to 4.2.6 left as a separate publish-time commit, per convention.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -12,6 +12,12 @@ console.log(`ReactFire v${version}`);
 
 const externals = [
   'react',
+  // Externalize the useSyncExternalStore shim so the consumer's bundler resolves it.
+  // Bundling this CJS module into the ESM output makes rolldown emit a dynamic require()
+  // that throws ("dynamic usage of require is not supported") when a reactfire data hook
+  // loads client-side in a Next.js App Router (strict ESM) context. Externalizing keeps
+  // the shim's React 16/17 fallback intact while removing the require from our dist.
+  'use-sync-external-store/shim',
   'firebase',
   'firebase/app',
   'firebase/firestore',


tyler-reitz:
@jhuleatt flagging this for an expedited **4.2.6**, since it is a regression in `latest`.

**What:** reactfire 4.2.4 and 4.2.5 crash on the client whenever a data hook renders. The bundled `use-sync-external-store/shim` becomes a dynamic `require()` that throws at module evaluation in any browser bundle without `require`.

**Blast radius (runtime-confirmed):** reproduced live in both a Next 16 App Router app (turbopack) and a Vite SPA on published 4.2.5, so it spans the two most common modern reactfire setups, not one bundler. Server prerender is unaffected (Node has `require`); the break is client-side. 4.2.3 is clean, so it came in with the build-tooling change around 4.2.4. No user reports yet is consistent with 4.2.4/4.2.5 being only a couple of days old.

**Fix (this PR):** one-line Vite `externals` change, non-breaking, React 16/17 fallback preserved. Verified the crash is gone (Next client chunk no longer carries the throw path; the Vite app renders `status: 'success'`). Armando is on for the technical review.

Since you are the manual publish path, this needs you for the 4.2.6 release once reviewed. Separately: this is the second dist-level regression to ship as a patch (after the 4.2.4 `.d.ts` change), so I think our release process needs a built-artifact diff gate before publish.

jhuleatt (approved):

Issue #765, opened 2026-07-24
Release process: add a built-artifact (dist/bundle) check to catch runtime regressions before publish


tyler-reitz:
### Motivation

#759 (the App Router / Vite client crash in 4.2.4 and 4.2.5) shipped because the build-tooling migration silently changed the **built ESM output**: `use-sync-external-store/shim` (CJS) started getting bundled into the ESM `dist`, producing a dynamic `require()` that throws in any browser bundle (`Calling \`require\` for "react" in an environment that doesn't expose the require function`). The source was unchanged; only the emitted artifact regressed, and nothing in the release flow compared the artifact before publish.

This is the **second dist-level regression to ship as a patch**: #749 already proposes a published `.d.ts` diff to catch the *type* side (the 4.2.4 `ObservableStatus` break). This issue covers the *runtime/bundle* side. Together they form a built-artifact diff gate.

### Proposed pre-publish checks

A concrete checklist for a pre-publish gate (each item derived from auditing the 4.2.6 release by hand). Items marked would have caught #759.

1. **No dynamic `require(` / CJS-interop shims in the ESM dist.** Grep `dist/index.js` for `\brequire\b`, `__require`, `createRequire`, `__commonJS`. *(Would have caught #759: `require` count went 0 at 4.2.3 to 3 at 4.2.4+.)*
2. **Externals are not inlined.** Only `rxfire` / `rxjs` / `tslib` should be bundled; `react`, `firebase/*`, `@firebase/*`, and `use-sync-external-store/shim` must stay external `import` specifiers. Catches accidental bundling that causes duplicate-instance bugs.
3. **Both entry points load.** `node --input-type=module -e "import('reactfire')"` and `node -e "require('reactfire')"` (in a fixture with `react` + `firebase` installed) must resolve without throwing. Catches missing/renamed files and broken imports.
4. **exports map integrity.** Every path referenced by `exports` / `main` / `module` exists in the packed tarball.
5. **Bundle-size delta vs previous `latest`.** `npm pack` the current `latest`, compare `dist` size; flag large jumps (a proxy for accidental inlining).
6. **Published `.d.ts` diff vs previous version.** Type-side counterpart, tracked in #749; catches the 4.2.4 `ObservableStatus` break class.
7. **Runtime smoke render in CI (strongest).** A minimal Next App Router (turbopack) and Vite app that renders a data hook against the packed build; fails on the #759 crash. This is the check that catches runtime regressions the static greps miss.

[...]

PR #766, opened 2026-07-27
ci: add built-artifact release gate (#765)

tyler-reitz:
## Why

Two dist-level regressions shipped as patch releases with nothing in the flow checking the built artifact:

- **4.2.4** turned `ObservableStatus<T>` into a discriminated union, breaking strict-TS consumers on a patch bump (#749).
- **4.2.4 / 4.2.5** inlined the CJS `use-sync-external-store/shim` into the ESM output, so `import('reactfire')` threw (#759, fixed by #760).

The source was fine both times. Only the emitted artifact regressed.

## What runs

One job, `Verify package`, against the same `reactfire.tgz` the `publish` job uploads, so it checks the exact bytes that ship.

| Tool | What it catches |
| --- | --- |
| `publint --strict` | manifest problems: missing `exports` conditions, bad paths, files referenced but not shipped |
| `attw` | type resolution across node10, node16 (CJS), node16 (ESM) and bundler |
| `size-limit` | entry-point gzip against a fixed budget |

No ignore rules. All four attw modes pass on their own now that #770 has landed.

## Manifest changes consumers will see

1. **A `types` condition on both `exports` branches.** Neither had one, so CJS consumers on node16 resolution got no declarations at all. Predates 4.2.4. This fix is what takes publint from exit 1 to "All good!". It needs the `.d.cts` copy: pointing the require branch at the plain `.d.ts` flips attw to a Masquerading-as-ESM failure.
2. **`sideEffects: false`**, which publint recommends. Verified not to change behavior: bundling fixtures with and without it through this repo's own vite/rollup toolchain produce byte-identical output, and the `globalThis` observable-cache write-backs survive minification.

Separately, `size-limit` was already a devDependency and already pointed at TSDX-era filenames the vite build stopped emitting, so `npm run size` was silently checking nothing.

## What this does not catch

Per review, `scripts/entry-load.mjs` and its job are removed. The gaps that leaves, stated plainly:

1. **#759's class is now uncovered.** publint and attw produce byte-identical output across 4.2.3, 4.2.4, 4.2.5 and 4.2.6. Both analyse package *structure*, and #759 was a `require` inlined into the *body* of a structurally valid ESM file. Actually loading the package was the only thing that caught it.
2. `size-limit`'s absolute budget will not catch a regression inside the current 3.7 kB of headroom. For scale, #759 moved the bundle +3.1%.
3. An undeclared import that happens to be a `react` or `firebase` transitive (say `@firebase/util`) is not flagged by publint. It would break a consumer whose tree does not hoist it.
4. A deleted `peerDependencies` block passes both tools. Neither asserts the manifest's shape beyond the paths it references.

## Scope

Covers #765 item 4, and item 5 within the caveat above. Items 1, 2 and 3 are declined per review rather than deferred, and item 7 is being split into its own issue, so **#765 closes as completed when this merges**. The `.d.ts` surface check for #749 is #777.


tyler-reitz:
@jhuleatt this is ready for review. All checks green. Summary of what needs you, roughly in order of how much it matters:

**1. One design decision, and it is the reason I have not merged this.**

#749 asks for the candidate's types to be diffed against **the last published version** on npm. This PR instead diffs against a **baseline committed in the repo**, updated deliberately via `npm run gate:accept`. Committed baseline surfaces the diff at PR time, needs no network, and cannot be skipped; but it can drift from npm if a release ever goes out without updating it. Diffing against npm cannot drift, and is closer to what the issue actually asks for.

My lean is the npm-published version, but I did not want to change it unilaterally, because **merging as-is auto-closes #749**. If we go the npm route I would rather downgrade that to a plain reference and keep the issue open. Happy either way, I just want the call to be yours since it is your issue.

**2. Two things needing admin, which I do not have.**

- Make **`Verify built artifact` a required status check on `main`**. Worth being precise: the gate already blocks **publishing** through the `needs: [test, verify-package]` edge, so that path is covered today. Making it required is about blocking **merges**.
- **Protect the `v5` branch.** `main` is protected, `v5` is not, and `v5` now holds real work (`aec5b83` forward-integration, `0b70a12` for #735). An accidental force-push or branch deletion would lose it. Ask is to mirror `main`'s protection onto `v5`, with one carve-out: **`Publish (NPM)` must not be a required check there**, because forward-integration heads are simultaneously `main`'s tip and inherit `main`'s own failing publish run. That is why #758 showed as `UNSTABLE` while still being mergeable.

**3. Heads-up on something this PR uncovered, no action needed from you here.**

This is the first PR to touch `.github/workflows/`, and the zizmor scan only runs when a workflow file changes, so it fired for the first time against a **pre-existing, repo-wide** condition. `main`'s `test.yaml` scores 34 findings (17 high, 8 medium). Two commits here clear it for that file. **`docs.yaml` is still unhardened** and will do the same thing to whoever edits it next. Worth its own pass at some point.

**4. Also correcting myself.** An earlier revision of this description claimed a working size guard would likely have caught the #759 shim inlining. That was wrong. Measured against the published tarballs, that regression moved the ESM entry only +3.1% gzip and the tarball +0.3%, so the ±10% band I originally used would have sailed straight past it. Tolerance is now ±2% and the description says plainly that `size` is a coarse packaging guard, not an inlining detector. Flagging it because it was the stated rationale for repairing the dead `size-limit` config.

**Scope note:** this implements #765 items 1, 2, 4, 5 plus #749. Items 3 and 7 (entry-point load test, runtime smoke render) are not included and are documented in `release-gate/README.md`, so **#765 stays open**. Item 7 is roughly a day on its own and adds real CI minutes and flake surface.

Separately, **#740** is still waiting on your semver call whenever you get a chance.

armando-navarro (dismissed):
Thanks Tyler. Reviewing the PR as it stands now, with the npm comparison and the new load test in. I verified the reworked machinery end to end and it holds up. Approving. The asks below are non-blocking on today's state, though the first two deserve attention before or right after merge.

### What I verified

- The 94 tests pass (69 gate, 25 load), and a live gate run against npm 4.2.6 passes with 0.0% entry deltas. Since both sides are now measured on the same machine, the compression-environment noise a committed baseline would carry is designed away.
- The acknowledgment lifecycle behaves: an unacknowledged type drift fails, `gate:accept` lets exactly that surface through, and any later edit invalidates the digest. The suite also pins that a new npm release expires an old acknowledgment.
- With the registry unreachable, the gate fails loudly after three attempts instead of skipping. A skip would be indistinguishable from a pass, so this is the right call, and the failure text says so.
- The load test catches the real thing: against the published 4.2.5 tarball it fails with the exact rolldown `require` crash on the ESM entry, and the current build loads both entries cleanly on React 18.
- The version rule blocks a type change shipped as a patch (the 4.2.4 case), stays quiet for CI's `-exp` version rewrite today while the majors match, and still applies to prereleases of a genuine bump.
- Pipeline wiring: `publish` needs the test job plus both verify jobs on both publishing triggers, the new job reuses the same pinned action commits, needs no secrets, and runs on fork PRs.

### The description no longer matches the PR

The body still describes the committed-baseline design. Five spots are now wrong:

- The "Decision needed before merge" section: the code has decided it.
- "Implements #765 items 1, 2, 4, 5": item 3 is in.
- The follow-ups bullet saying items 3 and 7 are not implemented.
- "38 tests": it is 69 plus 25 now.
- The new load-test job is not mentioned at all.

One more thought: since that decision had been explicitly flagged for Jeff, a short comment noting the switch and why would let his review start from the right frame rather than from the stale table.

On the footer specifically: your own description says that under the npm route "Closes #749" should become a plain reference with the issue left open, so that line wants the same update, with the final close-or-keep-open call staying with Jeff as you originally framed it. To be fair on the merits, the implementation does deliver #749's literal ask, so if Jeff prefers to let the merge close it, that is defensible too.

### The `latest` comparison will block every v4 release once v5 ships

This one is invisible today and painful later, and since the `v5` branch already exists I want it on the record now. I ran the version helpers against the future state where `5.0.0` holds npm `latest`:

- A v4 maintenance release `4.2.7` reads as a release candidate with `isMinorOrMajorBump` false, so the gate fails it as "a patch bump over 5.0.0".
- A v4 minor `4.3.0` is misjudged the same way, because a lower major can never register as a bump.
- Even a stamped canary build (`4.2.6-exp.<sha>`) reads as a release candidate against `5.0.0`, and every v4 PR fails the types diff against the v5 surface before the version rule even runs. So the blast radius is the whole v4 line's CI, not just release cuts.
- Acknowledging clears the types diff but not the version rule, so a contributor who hits this cannot get green without a code change. The change itself is small (the comparand is one constant), which is exactly why it is worth planning rather than discovering.

Fine to land as-is today (`latest` is 4.2.6, so nothing fires). My ask is just to file it with a clear "before 5.0.0 takes `latest`" milestone, or fix it here if you prefer. Two shapes that seem workable: make the compare target branch-aware (a per-branch dist-tag, or derive it from the candidate's own major), or make the version rule major-aware. Relatedly, once the gate lands on `v5`, every PR there will diff against the v4 surface and demand acknowledgment churn against a baseline that means nothing for that branch, so the same fix helps both.

### One gate:accept covers types and size together

`accepted.json` records the type digest and the size snapshot in one step, and the size check honors whatever the snapshot says:

- I built the ordinary case: a candidate with one benign added declaration file and a tarball 40% over published, acknowledged the normal way, passes with zero failures. The +40% shows up only as a log note.
- The committed file records the size as a raw byte count with no delta or percentage, so nothing in the acknowledgment diff itself flags that a size regression rode along with the type change.
- Bounded, to be fair: the primary inlining detectors do not consult the acknowledgment at all, so this only softens the size backstop.

Some cheap improvements:
- Have `gate:accept` print the size deltas it is about to bless, so whoever runs it sees what they are signing before committing (recording a readable delta in the file would go one better and put it in the PR diff too).
- Have it run the three absolute checks first and refuse to record on failure, which also keeps a broken build from becoming anyone's blessed snapshot.
- A related docs nit: the README's accepting-a-change section skips the build-first step its running-it section includes, so on a fresh clone the accept command errors until you build.

### Two carried notes on the checker itself

- The checks never look at `package.json`'s shape, only at whether referenced paths exist. A tarball with the `require` export condition removed, `typings` deleted, and `peerDependencies` deleted still passes everything (I re-ran this against the current head). An in-gate assertion on the extracted tarball's critical fields, excluding `version` since CI rewrites it, would close it.
- Nothing covers the wiring from a non-empty failures list to a non-zero exit. It works today (I checked all three exit codes), but a refactor could leave the gate exiting 0 forever with every test green. One child-process test against a crafted bad tarball would pin it.

### Small notes

- The load job has no retry or outage handling the way the gate's fetch does, and neither new job caches npm downloads, so every PR pays two uncached installs per React version. Fine to leave, worth knowing.
- The load fixture pins firebase to `^11.10.0`, which matches the repo's own devDependency today. When the repo moves to firebase 12, the fixture will quietly keep testing against 11 unless the constant moves with it. A one-line comment tying the two together would prevent the drift.
- `type-check` is not in `publish`'s needs list (pre-existing, not this PR): a type error in source would not block a canary.

If any of this reads wrong, say so and I will dig back in. The version-helper runs and the coupling fixture are one-liners to re-run.

tyler-reitz:
Thanks Armando, this was a genuinely useful review. Two of your asks are fixed and pushed (head is now `b50ede7`, CI green). One is dissolved rather than fixed, because the check it applied to has left this PR. The rest are open and I've said where each one is going.

## The `latest` comparison: fixed here rather than filed

You were right that this is invisible today and painful later, and right that it is worth planning rather than discovering. I took your first suggested shape, deriving the target from the candidate's own major, so a 4.x build now compares against `reactfire@^4`.

I went with derivation over a per-branch dist-tag for the same reason we dropped the committed baseline: a `v4` tag has to be maintained correctly on every publish, and since Jeff publishes by hand on a broken CI/CD path, it would drift. Deriving needs no upkeep.

When the candidate's major has nothing published (the v5 line before 5.0.0 ships), it falls back to the dist-tag and says so in the log. npm reports that case as `ETARGET`, which is distinct from the `E404` for a package that does not exist at all, so the two get different handling: no package stays a skip, no matching version falls back.

Verified on a runner rather than just locally. From the CI log on the current head:

```
release gate: reactfire@4.2.6-exp.9c0a62c (reactfire.tgz)
comparing against reactfire@4.2.6 (npm ^4)
```

That is CI's stamped version deriving `^4` correctly, which was the case I most wanted to see, since it is the one your analysis showed would take out every v4 pull request rather than just release cuts.

One related change worth flagging: `fetchPublished` now throws when a caller does not say what it is comparing, rather than defaulting to `latest`. Defaulting would reinstate exactly this bug, silently, in whichever call site forgot to pass the version. That turned out to matter within the hour, which I will come back to below.

## One `gate:accept` covering types and size: the types half has left this PR

Your ordinary case is a real hole, and I did not patch it. The types check moved out of this PR entirely, into the follow-up that carries a semantic API differ, so `gate:accept` here now covers size alone and there is nothing for a size regression to ride along on.

The reason for the move: the acknowledgment existed to make a human decide "additive or breaking". That decision is now made mechanically by a differ that asks `tsc` whether the new surface is assignable to the old one, so the ritual was buying a signature rather than a judgement, while reddening CI over a JSDoc edit and expiring every open PR's acknowledgment on each release.

Splitting it that way also means this PR now covers exactly #765, the runtime half, and the follow-up covers exactly #749, the type half. That is why the `Closes #749` footer is going away, which lands in roughly the same place you did, just for a different reason.

**This does mean part of what you verified no longer applies**, specifically the acknowledgment lifecycle: the type digest, its invalidation on a later edit, and its expiry on a new release. I would rather tell you than have you find it in the diff. The size half of that lifecycle is unchanged and your verification of it still holds. If you would rather re-look before this merges, say so and I will hold it.

Your other two suggestions on `gate:accept` still stand against the size-only version, and I have not done them: printing the deltas it is about to bless, and running the absolute checks first so a broken build cannot become a blessed snapshot. Both are worth doing. The README nit is fixed; the accepting section now includes the build step.

## A near miss worth recording

Removing the types check almost reintroduced the fail-open you verified earlier. `checkSize` was staying quiet on a failed registry fetch and relying on `checkTypes` having already failed the run for the same reason. Delete `checkTypes` and a registry failure becomes a silent skip across the whole gate.

Rather than move the coupling to the other check, `reportUnavailable` now dedupes on the report object, so both comparison checks report unconditionally and the first one wins. No check depends on another having run. Mutation-tested specifically, since it is invisible unless npm is unreachable.

The related catch: the differ was calling `fetchPublished` with no candidate version, and because that now throws rather than defaulting, it failed loudly on rebase instead of quietly comparing v4 builds against the v5 surface. That is the second time on this branch that a deliberate loud failure has paid for itself.

## Still open

- **The description.** Yours is the second correct read of it and it is now further out of date than when you wrote that, since the check count changed again. Rewriting before I mark this ready.
- **`package.json` shape assertion.** Agreed this is a real gap, and your repro is convincing. Going into the follow-up rather than growing this PR a third time.
- **Non-empty failures to non-zero exit.** Same, and I want the child-process test you describe rather than a unit test, for the reason you give.
- **Load job retry and npm caching, the firebase pin comment, `type-check` not in `publish`'s needs.** Noted, none done. The firebase pin one I would like to fix cheaply since the drift is silent.

Current state: 83 tests on this branch across the gate and load suites, no emulators needed, and the full workflow green on `b50ede7`. Still a draft until the description is rewritten.

tyler-reitz:
@armando-navarro flagging you directly, since [my reply above](https://github.com/FirebaseExtended/reactfire/pull/766#issuecomment-5122330619) asks you something and I did not mention you in it.

The part that needs you rather than just informing you: this is now marked ready and shows as approved off your 08:48 review, but that review predates the two commits that changed the PR's scope. The `types` check and its acknowledgment lifecycle, which you verified in detail, have moved out of this PR into the follow-up that classifies type changes mechanically. The size half of that lifecycle is unchanged and your verification of it still holds.

So: happy to hold the merge until you have re-looked. No rush at all, this can wait until you are back.

Fixed from your review in the meantime: the `latest` comparand (now derives the target from the candidate's own major, verified on a runner) and the shared `gate:accept`. Your remaining findings are all recorded in the follow-ups section of the description.

.github/workflows/test.yaml
```diff
@@ -13,15 +13,22 @@ on:
...
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
     name: Build
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
```
jhuleatt, line 27: Thanks for making the Zizmor changes, but please pull them out into their own PR and also change the docs workflow. That will keep this PR more focused

release-gate/README.md
```diff
@@ -0,0 +1,221 @@
...
+Implemented by [`scripts/release-gate.mjs`](../scripts/release-gate.mjs). It uses
+only node builtins plus `tar` and `npm pack`, so the CI job needs no `npm ci` and
+cannot itself be broken by a dependency change.
+
+This covers the **runtime** half of the release checks: does the built artifact
+still work. The **type** half, diffing the emitted `.d.ts` against the published
+release and applying semver to it, is
+[#749](https://github.com/FirebaseExtended/reactfire/issues/749) and is handled
+separately. It needs `typescript`, and keeping it out of this file is what lets
+the gate stay dependency-free.
```
jhuleatt: typescript is now covered by the normal test suite, correct?

test/entry-load.test.mjs
```diff
@@ -0,0 +1,223 @@
...
+    for (const name of EXPECTED_EXPORTS) expect(source).toContain(name);
+  });
+
+  it('reports a throw rather than letting the probe die silently', () => {
+    expect(probeSource('esm')).toContain("reason: 'threw'");
+    expect(probeSource('cjs')).toContain("reason: 'threw'");
+  });
+});
+
+describe('parseProbeOutput', () => {
```
jhuleatt: Some of these tests don't seem worth the extra lines of code. Could you please trim the test files down to core functionality?

release-gate/README.md
jhuleatt, line 1: Was this file meant to be included? It reads more like an agent conversation summary

jhuleatt (changes requested):
I appreciate the work, but need some convincing that this is worth the complexity it brings. `release-gate.mjs` alone is 700 lines, and I'm not confident I could debug it if it breaks down the line.

scripts/release-gate.mjs
```diff
@@ -0,0 +1,737 @@
...
+/**
+ * Built-artifact release gate.
+ *
+ * Verifies the packed tarball (the exact bytes CI publishes) rather than the
+ * working `dist/`, so what we check is what ships.
+ *
+ * Checks:
+ *   1. no CJS interop / dynamic `require` in the ESM entry      (issue #765, item 1)
+ *   2. externals stay external, nothing unexpected is inlined   (issue #765, item 2)
+ *   3. every path in `exports`/`main`/`module`/`typings` exists  (issue #765, item 4)
```
jhuleatt: 3 should be covered by the test suite, correct?

.github/workflows/test.yaml
tyler-reitz: Done, split out as #767, with `docs.yaml` included. `test.yaml` goes from 8 medium to 0, and `docs.yaml` clears entirely. 32 lines, no functional change to any job.

Worth flagging one thing: I applied the hardening to `main` directly rather than cherry-picking the two commits from here, because they were authored on top of the gate and carry the `verify-package` job with them. So this PR will still show the workflow changes until #767 lands and I rebase.

These commits will come out of this PR as part of the rework.

release-gate/README.md
tyler-reitz: Yes. #750's `tsc -p tsconfig.test.json` runs in CI today, and a faithful 4.2.4 repro (`data: T | undefined`, plus the non-null assert in `auth.tsx` so only consumers break rather than internals) fails it with 32 errors.

Only because unrelated tests read `.data` without narrowing though. Nothing is checking the type surface on purpose, so a test refactor could take the coverage with it. The api-extractor follow-up for #749 adds a real one.

scripts/release-gate.mjs
tyler-reitz: No. Check 3 there is `exports-map` (#765 item 4), and the test suite never loads the packed tarball, so nothing in it would notice a path in `exports` missing from the shipped files.

`publint --strict` covers it now, along with manifest problems we weren't checking at all. It caught a live one: neither `exports` branch had a `types` condition, so CJS consumers on node16 got no declarations. Fixed in this PR.

test/entry-load.test.mjs
tyler-reitz: Done. 25 tests down to 9, against a 220-line script.

One thing worth flagging: while trimming, I broke the `require()` half of the check on purpose to see if any test would notice. None did. Everything stayed green while the check was only really testing `import()`. The old 25 missed it too, so the trim exposed it rather than caused it. The 9th test is the one that catches it.

release-gate/README.md
tyler-reitz: You were right, it's gone. So is `release-gate.mjs` and its test file. Nothing replaced it: the two jobs are readable in `test.yaml`, and `entry-load.mjs` has a short header saying why it exists alongside `publint` and `attw`.

tyler-reitz:
Reworked this in place rather than opening a new PR, so your comments stay attached. It's a different PR now: `release-gate.mjs` (737 lines), its tests (666) and the README are gone, replaced by `publint`, `attw`, `size-limit` and one 220-line load script. 7 files, +689/-10.

Description is rewritten and all five of your comments have replies. Three are on files that no longer exist, so they'll show as outdated.

Based on #767 so the diff here is only this PR's own work. It'll retarget to `main` when that lands.

armando-navarro (approved):
Thanks Tyler. This is my re-look, so consider the hold answered: approving this shape. Replacing the bespoke checker with off-the-shelf tooling after Jeff's pushback reads right to me, and the description now says plainly what the simpler approach gives up, which made this round easy to verify.

### What I verified

- `publint --strict` is clean on the build, and attw is green on all four resolution modes with exactly the one suppressed rule. Without the config, the only reported problem class is the internal-resolution error, so the suppression hides nothing else.
- Your central claim reproduces exactly: publint and attw output is identical on the published 4.2.5 and 4.2.6 tarballs, so neither can see the #759 class, while `entry-load` fails 4.2.5 with the real crash message and passes the current build.
- The `types`-condition fix is real, and the `.d.cts` copy is not just correct but necessary: pointing the require branch at the plain `.d.ts` flips attw to a Masquerading-as-ESM failure, which doubles as proof the attw check is load-bearing.
- The workflow wiring holds: `publish` needs all three upstream jobs, the unpack step means publint lints the shipped manifest and size-limit reads the shipped `dist/`, attw checks the tarball bytes directly, pins and fork-PR safety carried over from #767, and the lockfile delta is purely publint and attw.
- The 9 load tests pass, and the ones you kept are the right ones (each pins a way the check could silently stop detecting).

### The load probe covers 4 of the 10 submodules

Small find: the probe's failure condition is narrower than its comment claims, and the gap has a realistic mechanism.

- `EXPECTED_EXPORTS` canaries one export each from 4 submodules (firestore, auth, firebaseApp, the core hook), while `index.ts` re-exports from 10, so a drop confined to storage, functions, database, performance, remote-config, or sdk passes the probe.
- The source-level jobs cannot catch that case: type-check and the test suite run against `src`, and only this probe loads the packed artifact, so a bundling regression that drops artifact exports leaves everything else green.
- The script's comment says it catches "an empty or partial build", which is broader than what it asserts, and with the type-surface check deferred to the api-extractor follow-up, nothing else catches export drift in the meantime.

Cheapest fix: widen the canary list to one export per submodule, which is self-documenting. If you would rather leave it to the follow-up, scoping the comment down to "an empty build or one missing a core export" would make the claim match the behavior.

### One manifest change worth a line in the description

`sideEffects: false` is in the commit message but not the PR body, and since it changes how consumers' bundlers treat the package I went at it hard. It held up:

- publint itself recommends the flag, so it belongs to this PR's story rather than beside it.
- Bundling fixtures with and without the flag through this repo's own vite/rollup toolchain produced byte-identical output.
- The `globalThis` cache write-backs survive minification, so the shared-cache behavior consumers rely on is intact.

So it is safe in practice. A line in the description alongside the `types`-condition fix would make the consumer-facing change as visible as the CI change.

### The suppression needs its issue to exist

The `.attw.json` ignore is fine exactly as long as it is temporary, and right now nothing anchors that.

- The ignore is rule-global, since attw offers nothing narrower, so any future error of that class hides under it too.
- The description says the extensionless-imports bug "gets its own issue", but I could not find one filed yet.

Filing it at or around merge would give "delete the ignore rule" a home as that issue's acceptance test, and keep the suppression from quietly becoming permanent.

### Small notes

- The probe's missing-export filter is only ever tested through a mocked runner, and a green CI run never exercises that branch, so a broken filter would not be noticed. A one-line unit test on `probeSource`'s generated check would pin it.
- Two narrow classes the old checker covered have no successor, and one line each in "what this gives up" would make the retirement read as deliberate: an accidental undeclared import that happens to be a firebase or react transitive (say `@firebase/util`) loads fine in the fixture and publint does not flag undeclared bare imports, and a deleted `peerDependencies` block passes both publint and attw (I checked both).
- Worth a CI re-check after this retargets to main once #767 lands: main has already moved under the branch (the #768 re2 bump), and it touches the same lockfile this PR edits.

If any of this reads wrong, say so and I will dig back in. The 4.2.5-vs-4.2.6 tool-output comparison and the masquerade flip are quick to re-run.

tyler-reitz:
Thanks @armando-navarro. Three of your four asks are in as of `774dde2`. Nothing here contradicts what you verified, so the approval should still stand, but the head moved after it so here is what changed.

**The probe now canaries one export per submodule.** Confirmed your count: 4 canaries against the 10 modules `index.ts` re-exports, so a drop confined to storage, functions, database, performance, remote-config or sdk passed. `EXPECTED_EXPORTS` is now one entry per submodule with the module named in a trailing comment. I also rewrote the comment above it, since your second point was right that "an empty or partial build" claimed more than the assertion delivered. Verified: the 9 load tests pass, all ten names appear in the emitted `dist/index.js`, and its export count is 74, matching what CI reports.

**The suppression has an anchor: #769.** Measured fresh for the issue rather than reusing the earlier figure: 16 extensionless relative imports across 10 emitted `.d.ts` files, plus 6 bare `'./'` specifiers that node16 also rejects. The acceptance test is stated as you framed it, delete the `ignoreRules` entry and keep `Verify package` green. The description here now points at it.

**`sideEffects: false` is in the description now**, alongside the `types`-condition fix, with your verification result recorded (byte-identical bundles with and without, `globalThis` cache write-backs surviving minification).

On the small notes: the `probeSource` unit test, the firebase-fixture pin comment, and the "what this gives up" lines for the two retired classes are worth doing and are not in this push. Say the word if you would rather they land before merge than after.

CI is green on `774dde2`. One thing to flag: `Test Node.js 24` failed on the first run at `test/firestore.test.tsx:180` (`waitFor` timing out with `'loading'` instead of `'success'`), and passed on a re-run with no code change. That is the recurring flake in that file, and this PR touches no source at all.

tyler-reitz:
@armando-navarro heads up: this branch changed after your approval, so it is worth a fresh look. Two things.

**1. Rebased onto `main`.** #767 merged, so the base auto-retargeted and the stack is gone. The diff is now **6 files instead of 8**: `.attw.json` is deleted (below), `docs.yaml` dropped out because it belonged to #767, and `test.yaml` went from +86/-13 to +61/-1. No content of this PR was removed, it just stopped carrying #767's changes.

**2. The attw suppression is gone.** #770 merged first and gave the emitted `.d.ts` explicit `.js` extensions, so `internal-resolution-error` no longer fires and the ignore rule was dead weight. Verified against a packed tarball with no config file present:

```
node10            🟢
node16 (from CJS) 🟢 (CJS)
node16 (from ESM) 🟢 (ESM)
bundler           🟢
No problems found 🌟
```

`attw` exits 0. I deleted `.attw.json` outright rather than emptying it, since the ignore rule was its only content and `lint:types` is a bare `attw` invocation.

This also closes out the `.attw.json` anchor you asked for: #769 is fixed and closed, so there is no longer a suppression that needs an anchor.

One thing worth naming: had #766 merged before #770, the suppression would have shipped and then needed a follow-up to remove. Keeping the types fix as its own PR is what avoided that.

This is still blocked on @jhuleatt's `CHANGES_REQUESTED`, which a new approval does not clear.

jhuleatt (changes requested):
I like the addition of attw and publint. Let's stick with those, and get rid of scripts/entry-load. If we continue to ship regressions that would have been caught by an additional check like scripts/entry-load, we can reconsider.

tyler-reitz:
Will do, removing `scripts/entry-load` and keeping publint and attw.

One note for the record. Armando verified on Aug 3 that publint and attw produce identical output on the published 4.2.5 and 4.2.6 tarballs, so neither can see the #759 crash, while `entry-load` fails 4.2.5 with the actual runtime error and passes 4.2.6. So it's the regression we shipped in July rather than a hypothetical one. #777's api-extractor check doesn't close the gap either, since it reads declarations instead of loading the packed artifact.

Your call, stripping it now.

PR #770, opened 2026-08-03
fix(types): use explicit .js extensions in relative imports

### LINKED ISSUE #769: types: emitted .d.ts use extensionless relative imports, which node16 resolution rejects
## Problem

The emitted `.d.ts` files use extensionless relative specifiers (`from './useObservable'`), which `node16`/`nodenext` module resolution rejects. TypeScript consumers on those settings get resolution errors reading our declarations.

Measured against the current build (`npx tsc && npx vite build`):

- **16 extensionless relative imports across 10 emitted `.d.ts` files**, plus **6 bare directory specifiers (`from './'`)**, which `node16` also rejects.
- The most repeated one is `from './useObservable'` (5 occurrences); every submodule entry (`./auth`, `./database`, `./firestore`, `./functions`, `./performance`, `./remote-config`, `./sdk`, `./storage`) contributes one from `src/index.ts`.

The cause is in source, not in the build: `src/index.ts` re-exports with `export * from './auth'` and friends, and `tsc` emits specifiers unchanged. So the fix belongs in `src`, which is why it was kept out of #766.

## Why this is filed now

#766 adds `@arethetypeswrong/cli` to CI. Without a suppression it fails on this, so the PR ships:

```json
// .attw.json
{ "ignoreRules": ["internal-resolution-error"] }
```

`attw` offers nothing narrower than a rule-global ignore, so that entry also hides **any future error of the same class**, not just this one. The suppression is only defensible while it is temporary, and right now nothing tracks that.

## Acceptance test

**Delete the `ignoreRules` entry from `.attw.json` and have the `Verify package` job stay green.** That is the whole test. If the ignore cannot be removed, the fix is incomplete.

## Notes

- Fixing this means adding explicit extensions to relative specifiers in `src` (`./useObservable.js` style, which is what `node16` wants even from `.ts` sources), and giving `./` re-exports a concrete file.
- Worth checking whether `moduleResolution` in `tsconfig.json` should move in the same change, so the compiler enforces this rather than leaving it to `attw` in CI.
- Scope is declarations only. The runtime bundles resolve fine today, so this is not a user-visible runtime break, only a types-resolution one.

Evidence for the `attw` behavior (that the rule fires on exactly this class, and that suppressing it hides nothing else) was recorded when #766 was built; the counts above are from a fresh build today.

tyler-reitz:
## Summary

Fixes #769. Adds explicit `.js` extensions to relative import specifiers in `src`, so the emitted `.d.ts` stop carrying extensionless paths that `node16` / `nodenext` resolution rejects.

Source-only change, 10 files, all mechanical. `moduleResolution: bundler` accepts both forms, so nothing changes for the build or for bundler consumers.

## The bug

TypeScript emits relative specifiers unchanged, so `export * from './auth'` in `src/index.ts` becomes `from './auth'` in `dist/index.d.ts`. Under `node16` that is not a valid specifier, and a consumer on `"moduleResolution": "node16"` gets resolution errors reading our declarations.

Two shapes were involved:

- **16 extensionless relative imports across 10 emitted `.d.ts` files** (`from './useObservable'`)
- **6 bare directory specifiers**, `from './'` and `from '.'`, in `auth`, `database`, `firestore`, `functions`, `remote-config`, `sdk`, `storage` and `useObservable`. These resolve to the barrel and now point at `./index.js` explicitly

The runtime bundles resolve fine, which is why nothing in the build caught it.

## Verification

`@arethetypeswrong/cli` against the packed build, **with no ignore rules**:

| | before | after |
| --- | --- | --- |
| node10 | pass | pass |
| node16 (from CJS) | no types | no types |
| node16 (from ESM) | **internal resolution error** | **pass** |
| bundler | pass | pass |

The remaining `node16 (from CJS)` failure is a different bug: no `types` condition on either `exports` branch, fixed in #766. **Applying both, all four modes pass with no suppression**, which is the acceptance test #769 asks for.

Also confirmed: `tsc --noEmit` clean on both `tsconfig.json` and `tsconfig.test.json`, `vite build` succeeds, and zero extensionless relative specifiers remain in the emitted declarations.

## Relationship to #766

#766 adds `attw` to CI and currently suppresses this via `.attw.json`:

```json
{ "ignoreRules": ["internal-resolution-error"] }
```

That ignore is rule-global, so it hides any future error of the class. **If this lands first, #766 can drop the ignore before it merges** and the suppression never ships. If #766 lands first, deleting the entry becomes this issue's acceptance test instead. Either order works; this one is tidier.

## Test plan

- [ ] `npm test`
- [ ] `npx tsc --noEmit` on both tsconfigs
- [ ] `attw` on the packed build shows no internal resolution error

### DIFF
--- a/src/SuspenseSubject.ts
+++ b/src/SuspenseSubject.ts
@@ -1,6 +1,6 @@
 import { empty, Observable, Subject, Subscriber, Subscription } from 'rxjs';
 import { catchError, shareReplay, tap } from 'rxjs/operators';
-import { ObservableStatus } from './useObservable';
+import { ObservableStatus } from './useObservable.js';
 
 export class SuspenseSubject<T> extends Subject<T> {
   private _value: T | undefined;
--- a/src/auth.tsx
+++ b/src/auth.tsx
@@ -1,9 +1,9 @@
 import * as React from 'react';
 import { user } from 'rxfire/auth';
-import { preloadObservable, ReactFireOptions, useAuth, useObservable, ObservableStatus, ReactFireError } from './';
+import { preloadObservable, ReactFireOptions, useAuth, useObservable, ObservableStatus, ReactFireError } from './index.js';
 import { from, of, defer } from 'rxjs';
 import { map, switchMap } from 'rxjs/operators';
-import { useSuspenseEnabledFromConfigAndContext } from './firebaseApp';
+import { useSuspenseEnabledFromConfigAndContext } from './firebaseApp.js';
 
 import type { Auth, User, IdTokenResult } from 'firebase/auth';
 type Claims = IdTokenResult['claims'];
--- a/src/database.tsx
+++ b/src/database.tsx
@@ -1,5 +1,5 @@
 import { list, object, QueryChange, listVal, objectVal } from 'rxfire/database';
-import { ReactFireOptions, useObservable, checkIdField, ObservableStatus, ReactFireGlobals } from './';
+import { ReactFireOptions, useObservable, checkIdField, ObservableStatus, ReactFireGlobals } from './index.js';
 
 import type { Query as DatabaseQuery, DatabaseReference } from 'firebase/database';
 
--- a/src/firestore.tsx
+++ b/src/firestore.tsx
@@ -1,6 +1,6 @@
 import { collectionData, doc, docData, fromRef } from 'rxfire/firestore';
-import { ReactFireOptions, useObservable, checkIdField, ReactFireGlobals } from './';
-import { preloadObservable, ObservableStatus } from './useObservable';
+import { ReactFireOptions, useObservable, checkIdField, ReactFireGlobals } from './index.js';
+import { preloadObservable, ObservableStatus } from './useObservable.js';
 import { first } from 'rxjs/operators';
 
 import { Query as FirestoreQuery, QuerySnapshot, DocumentReference, queryEqual, DocumentData, DocumentSnapshot } from 'firebase/firestore';
--- a/src/functions.tsx
+++ b/src/functions.tsx
@@ -1,7 +1,7 @@
 import { httpsCallable as rxHttpsCallable } from 'rxfire/functions';
 import { defer } from 'rxjs';
-import { ReactFireOptions, useObservable, ObservableStatus } from './';
-import { useFunctions } from '.';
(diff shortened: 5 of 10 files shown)


armando-navarro (approved):
Thanks Tyler, and thanks for filing #769 out of the #766 discussion. This is the fix I was hoping for, and it approves cleanly.

### What I verified

I reproduced the problem on `main` and confirmed the fix against a fresh build, using `attw` with no ignore rules:

- On `main`, node16-from-ESM resolves the declarations to an internal resolution error, and the emitted `.d.ts` carry 16 extensionless relative specifiers.
- On this branch, node16-from-ESM goes green and zero extensionless relative specifiers remain in the emitted declarations.
- node16-from-CJS stays on "no types", which is the missing `types` condition and lands with #766, exactly as your description says.
- `tsc --noEmit` is clean on both `tsconfig.json` and `tsconfig.test.json`, and `vite build` succeeds.
- The runtime bundles (`dist/index.js`, `dist/index.umd.cjs`) carry no relative specifiers, so this is resolved away at build time and nothing changes for runtime or for bundler consumers.

I also reverted a single specifier back to extensionless as a check, and the internal resolution error came straight back, so the fix is doing real work across all 29 edits, not riding on one.

### One thing worth a follow-up (non-blocking)

The fix is correct, but I do not think it is self-enforcing yet. `tsconfig.json` stays on `moduleResolution: bundler`, which accepts both the extensionless and the `.js` forms, so `tsc` will stay green if an extensionless relative import creeps back in later. And there is no `attw`/`publint` job in CI today, so nothing catches a regression until #766 adds one.

Your own note on #769 floated moving `moduleResolution` in the same change so the compiler enforces this. Would you want to do that here, or add a lint rule for relative-import extensions, so this cannot silently regress before #766 lands? I have not checked whether switching resolution builds cleanly (it can have knock-on effects), so I am asking rather than recommending.

### On the #766 suppression

Confirming the coordination in your description: with this branch plus #766's `types` condition, all four `attw` modes pass with no ignore rules. So whichever order these land, #766 can drop the `.attw.json` `ignoreRules` entry and the suppression never ships permanently. Landing this one first is the tidier path, as you say.

### Small note

`src/functions.tsx` now has two separate imports that both read `from './index.js'` (they were `from './'` and `from '.'` before). Totally valid, just foldable into one line if you feel like it.

If any of this reads wrong, say so and I will take another look.

Issue #776, opened 2026-08-04
CI: test/firestore.test.tsx flakes intermittently on both Node 22 and Node 24


tyler-reitz:
`test/firestore.test.tsx` fails intermittently in CI and passes on re-run with no code change. It has now fired five times in three days, every time on a PR that cannot plausibly cause it, including a lockfile-only dependency bump and a README-only docs change.

Every occurrence costs a full CI cycle and a manual re-run, and it trains reviewers to ignore red runs.

> ### ⚠️ Two corrections, 2026-08-05
>
> This issue previously said the flake was **Node 24 only** and that the `RESOURCE_EXHAUSTED` error was **ruled out**. Both were wrong, and the [earlier investigation comment](https://github.com/FirebaseExtended/reactfire/issues/776#issuecomment-5185351270) is superseded on those two points.
>
> 1. **It is not Node-24-specific.** On #778, a README-only diff, `Test Node.js 22` failed **twice consecutively** while `Test Node.js 24` passed in the same runs. The earlier "Node 22 passes every time" claim rested on a local sample of 6 runs, which was underpowered rather than clean.
> 2. **`RESOURCE_EXHAUSTED` is not ruled out.** It is absent from every *local* reproduction, which is all the evidence supported. In *CI* it is present, including on the #778 failures. **This raises the possibility that the local repro and the CI failure are not the same bug**, which matters because the fix below was measured against the local one.

## Signature

```
AssertionError: expected 'loading' to deeply equal 'success'
Expected: "success"
Received: "loading"
```

In CI, alongside it on the same run:

```
@firebase/firestore: Firestore (11.10.0): GrpcConnection RPC 'Listen' stream error.
Code: 8 Message: 8 RESOURCE_EXHAUSTED: Received message larger than max (795107700 vs 4194304)
```

A ~795 MB (and in an earlier run ~3.5 GB) claimed frame against a 4 MB cap is a **gRPC framing desync**: the reader is interpreting payload bytes as a length header. That is a `@grpc/grpc-js`-layer failure, not an application one.

Locally the failure looks different: `waitFor` hitting its 1000ms default, at 1058ms and 1142ms, plus one 120s hang, with **no** gRPC error at all.

## Occurrences

| Date | PR | What it touched | Which Node |
| --- | --- | --- | --- |
| 2026-08-05 | #778 | `README.md` only | **22** (twice in a row; 24 passed) |
[...]

PR #778, opened 2026-08-04
docs: replace the concurrent mode note with a Suspense section

### LINKED ISSUE #756: docs: remove note about "extra experimental features"
In https://github.com/FirebaseExtended/reactfire#extra-experimental-concurrent-mode-features, we say "These features are marked as extra experimental because they use experimental React features that will not be stable until sometime after React 18 is released."

I think the whole section should be removed, with a new bullet point added to https://github.com/FirebaseExtended/reactfire#what-is-reactfire mentioning Suspense support.

tyler-reitz:
Closes #756.

The "Extra Experimental concurrent mode features" section rests on a claim that is no longer true. It says these features "will not be stable until sometime after React 18 is released". React 18 shipped in 2022, and concurrent mode was abandoned as a concept rather than stabilised, so a reader today is told to follow a setup guide for a React feature that does not exist.

All three of its `reactjs.org/docs/concurrent-mode-*` links are dead pages. This PR removes all three from the README. One more survives in a comment in `example/index.tsx`, left for a separate follow-up so this stays docs-scoped.

## What changed

- Removed the "Extra Experimental concurrent mode features" heading and its stale premise paragraph.
- Added a `<Suspense>` bullet to **What is ReactFire?**, as the issue asked.
- Rehomed the remaining content as a plain `## Suspense` section.
- Added that suspense is **off by default**, which the old text never stated. Verified against `src/firebaseApp.tsx` (`suspense ?? false`, lines 57 and 66).
- Described what `<SuspenseWithPerf />` actually does. It measures with the browser User Timing API, not the Firebase Performance SDK. See below.

## What was kept, and why

The issue asks for the whole section to be removed. Two things were kept:

- the `suspense={true}` snippet, which is the only place the repo documents how to turn suspense on
- the `<SuspenseWithPerf />` mention, which was the only prose reference to that component anywhere outside the generated API report

The stale part was the "extra experimental / concurrent mode" framing and its dead links, not the usage documentation, so that framing is what this removes.

## The Performance Monitoring claim, corrected in review

The first revision carried the old text's claim that `<SuspenseWithPerf />` "instruments load times with Performance Monitoring", linking the Firebase perf-mon docs. Armando checked it against the source and it is wrong: `src/performance.tsx` uses only `performance.mark` / `performance.measure` from the browser's User Timing API, and still carries a `// TODO: Should this import firebase/performance?`. Nothing there touches the Firebase Performance SDK.

The claim is inherited rather than introduced here, but this PR rewrites that sentence anyway, so it is fixed rather than carried forward. The line now describes the User Timing behaviour and links MDN.

## Notes

- Docs-only. No source change, so no reference docs regeneration is needed.
- This section gets rewritten again at V5, when the suspend path moves to `use()`. It was kept minimal for that reason.
- If `Test Node.js 24` goes red here, it is very likely #776 rather than anything in this diff, given this PR touches only `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ Firebase.
   auth state, realtime data, and all other Firebase SDK events. Plus, they automatically unsubscribe when your component unmounts.
 - **Access Firebase libraries from any component** - Need the Firestore SDK? `useFirestore`. Remote Config? `useRemoteConfig`.
 - **Safely configure Firebase libraries** - Libraries like Firestore and Remote Config require settings like `enablePersistence` to be set before any data fetches are made. This can be tough to support in React's world of re-renders. ReactFire gives you `useInitFirestore` and `useInitRemoteConfig` hooks that guarantee they're set before anything else.
+- **Optional `<Suspense>` support** - Hand loading states to React instead of checking a status yourself. Off by default, opt in with the `suspense` prop. See [Suspense](#suspense) below.
 
 ## Platform support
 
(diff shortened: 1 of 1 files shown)


armando-navarro (approved):
Thanks Tyler. This reads a lot better than the old "extra experimental" framing, and getting rid of the dead concurrent-mode links is overdue. I checked the changed claims against the source and they hold up, so I am comfortable approving. Two small things, neither blocking.

### What I verified

- "Off by default" is right: `FirebaseAppProvider` provides the suspense context as `suspense ?? false` (`src/firebaseApp.tsx:57`), and `useIsSuspenseEnabled` returns the same default (`:66`). Good to finally have this written down.
- "Hooks can throw promises for `<Suspense>` to catch" matches the code: `useObservable` throws `observable.firstEmission` (a promise) when suspense is on and there is no data yet (`src/useObservable.ts:80-81`).
- The `[Suspense](#suspense)` bullet link resolves, since `## Suspense` is the only heading of that name, and the new `react.dev` and `example/withSuspense` links are live.

### One optional wording note

The new line says `<SuspenseWithPerf />` "instruments load times with Performance Monitoring" and links the Firebase Performance Monitoring docs. Reading `src/performance.tsx`, the component measures with the browser User Timing API (`performance.mark` / `performance.measure`) and carries a `// TODO: Should this import firebase/performance?`, so it is not actually wired to the Firebase performance SDK. The old text linked the same page, so this is inherited rather than anything you introduced. Since you are already rewriting that sentence, it might be a natural moment to soften "Performance Monitoring" or drop the perf-mon link, but I am happy either way.

### Minor, out of scope

Your description mentions removing "the last of those links from the repo". There is still one `reactjs.org/docs/concurrent-mode-adoption.html` link in a comment in `example/index.tsx`. The README is the right scope for this PR, so no need to touch it here, just flagging in case you want a quick follow-up to finish them off.

Approving. If I have misread any of the above, point me at it and I will take another look.

tyler-reitz:
Thanks Armando, both findings were right and both are now addressed.

**Performance Monitoring.** You are correct and I took the fix rather than leaving it. `src/performance.tsx` measures with `performance.mark` / `performance.measure` only, and the `// TODO: Should this import firebase/performance?` is still sitting there, so the perf-mon link pointed at a product the component never calls. Since I was rewriting that sentence anyway, carrying the claim forward would have meant knowingly shipping it. The line now reads:

> `<SuspenseWithPerf />` does the same and also measures how long the fallback was shown, using the browser's [User Timing API](https://developer.mozilla.org/en-US/docs/Web/API/Performance_API/User_timing).

That is in `e1dba04`, a second commit rather than a force-push, so your approval stays attached to what you reviewed.

**The remaining link.** Also correct, and the PR body was the actual problem: it claimed this removed "the last of those links from the repo", which is false while `example/index.tsx:8` still points at `concurrent-mode-adoption.html`. I have corrected the body to say the README, and noted the example as a follow-up. Agreed it does not belong in a docs PR.

Merging on your approval under the patch-level rule. The `firebase/performance` TODO is a real open question about whether that component should use the SDK at all, but that is a behaviour decision for V5, not a README fix.

PR #779, opened 2026-08-05
fix(ssr): add getServerSnapshot to useObservable's useSyncExternalStore

### LINKED ISSUE #748: useObservable: add getServerSnapshot to useSyncExternalStore for SSR
\`useObservable\` calls \`useSyncExternalStore(subscribe, getSnapshot)\` with only two arguments. React 18 requires a third \`getServerSnapshot\` for server rendering. Without it, React throws during SSR and Next.js App Router falls back to client rendering for any subtree using a reactfire hook.

The fix is to pass a \`getServerSnapshot\` that returns the same seeded \`immutableStatus\` the client snapshot returns. The server snapshot must be stable across server/client to avoid hydration mismatches, so \`initialData\` seeding is the natural tie-in.

Flagged in the arch review as a V5 "just do it" item: not a spike, not blocked on any API decision.

tyler-reitz:
Fixes #748.

`useObservable` called `useSyncExternalStore` with two arguments. React requires a third, `getServerSnapshot`, whenever the tree is server rendered or hydrated. Without it React throws `Missing getServerSnapshot, which is required for server-rendered content` and the surrounding subtree silently falls back to client rendering, which is why a Next.js App Router page using any reactfire hook loses SSR for that subtree.

## The part worth reviewing

The issue proposes returning "the same seeded `immutableStatus` the client snapshot returns". **That version is unsafe and this PR deliberately does not do it.**

`preloadedObservables` is a `Map` on `globalThis`, keyed only by `observableId`. In a browser that is one user's cache. On a server it is shared by every concurrent request, so a `getServerSnapshot` that read `observable.immutableStatus` would render data request A fetched into request B's HTML whenever both touch the same path.

That leak is unreachable today only because SSR throws before it can happen. So fixing the crash the obvious way would trade a crash for a cross-request data disclosure, which is the worse of the two.

This implementation reads **only `config`** for anything data-bearing, and `config` arrives from the caller on the current render, so it is per-request. With `initialData` it reports `success` and that value; without it, `loading`. (`firstValuePromise` does read the shared observable, but it is a `Promise<void>` that resolves without a value, so nothing crosses through it.) The result is held in a ref so the value is stable across renders, which is what React's "The result of getServerSnapshot should be cached" check wants.

## Verification

Six tests under a new `Server rendering` block, and both mutations were run rather than assumed:

- **Drop the third argument:** all six fail, with React's own `Missing getServerSnapshot` error. So the tests are load-bearing rather than decorative.
- **Return `observable.immutableStatus` instead** (the straightforward implementation): **4 of 6 pass and 2 fail**, `does not leak a cached value from another request into the server snapshot` and `prefers the callers initialData over a value already in the shared cache`. Those two are the tests that earn their place, and they are why the constraint above is written down in the code rather than left to reviewer memory.

`tsc` passes on both `tsconfig.json` and `tsconfig.test.json`; `useObservable` is 20/20; eslint reports 0 errors on both changed files.

## Scope: this does not make SSR work, and the title should not be read that way

**In non-suspense mode without `initialData`, a server render produces `loading` placeholders**, not data. That is inherent to the model rather than a gap in it, for the reason above: `observableId` alone cannot distinguish this request's data from another request's. Real server-side data needs per-request cache scoping, which is a separate piece of work.

**Suspense mode reaches `getServerSnapshot` too, and its server behavior does change.** The suspend decision is `hasData = observable.hasValue || hasInitialData`, so whenever the shared cache already holds a value for the id, the guard at `src/useObservable.ts:80` is skipped and the hook runs. A `SuspenseSubject` subscribes in its constructor, so even a first render that crashes seeds that entry: on a long-lived server process the already-cached case is the steady state, not an edge.

Measured with the same probe on both branches:

| `suspense: true` | `main` | this branch |
| --- | --- | --- |
| Cold cache, `renderToString` | `A component suspended while responding to synchronous input` | same |
| Cold cache, streaming | hangs, no completion | same |
| **Value cached for the id**, `renderToString` | `Missing getServerSnapshot` | renders `loading:undefined` |
| **Value cached for the id**, streaming | `onShellError`, empty output | completes with `loading:undefined` |
| Value cached, consumer dereferences `data` | `Missing getServerSnapshot` | `TypeError: Cannot read properties of undefined` |

**The cost is diagnosability.** A named library error becomes silent `loading` HTML, or a `TypeError` inside the consumer's own component, since suspense mode's contract is that the hook does not return until data exists and callers dereference accordingly. Not fixed here and not fixable without per-request scoping, but worth stating rather than leaving as "suspense mode cannot server-render".

## Notes

- **The `use-sync-external-store/shim` is a red herring for React 18 and 19.** Its own implementation ignores `getServerSnapshot` on purpose, but it resolves to `React.useSyncExternalStore` whenever React exposes it, so the third argument reaches React's real implementation. Reading only the shim would suggest this fix cannot work. **On 16 and 17, which the peer range still allows, it does not:** the shim's server path returns `getSnapshot()`, so the cached value reaches the markup there. Unchanged from `main` and out of scope here, and the source comment is scoped accordingly.
- **No reference docs regeneration**, since no exported symbol changed.
- The `globalThis` cache itself is unfixed. This PR contains the hazard at the one place it would otherwise become reachable, and nowhere else.
- The server snapshot literal carries no `as` cast, so `tsc` fails with TS2741 if a required field such as `firstValuePromise` goes missing.
- One `eslint-disable` for `react-hooks/exhaustive-deps`, with the reason in a comment above it: callers routinely pass a fresh `config` literal each render, and the ref means the value is computed once per component instance anyway.

### DIFF
--- a/src/useObservable.ts
+++ b/src/useObservable.ts
@@ -109,15 +109,26 @@ export function useObservable<T = unknown>(observableId: string, source: Observa
   // required for server-rendered content" and the surrounding subtree falls back to
   // client rendering.
   //
+  // This applies on React 18 and up. Below that, `use-sync-external-store/shim` ignores the
+  // third argument on the server as well as the client, so both shapes render identically
+  // and nothing here regresses. It also does nothing for it.
+  //
   // This deliberately does NOT return `observable.immutableStatus` the way `getSnapshot`
   // does. `preloadedObservables` is a `globalThis` cache keyed only by `observableId`, so
   // on a server it is shared by every concurrent request. Seeding the server snapshot from
-  // it would let one request render data another request fetched for the same path. Only
-  // `config` is safe to read here, because it comes from the caller on this render.
+  // it would let one request render data another request fetched for the same path.
+  //
+  // So no field below carries data across requests: `status`, `hasEmitted` and `data` come
+  // from `config`, which is the caller's own input on this render. `firstValuePromise` does
+  // read the shared `observable`, but it is a `Promise<void>` that resolves without a value,
+  // so it discloses nothing. It is not optional on `ObservableStatus`, and the
+  // `as ObservableStatus<T>` cast below means neither `tsc` nor the tests would notice if it
+  // were dropped: a caller doing `status.firstValuePromise.then(...)` on the server would
+  // just throw.
   //
   // The result is memoized per component instance because React compares the value it
   // returns across renders, and a fresh object each time is what triggers the
-  // "The result of getSnapshot should be cached" error.
+  // "The result of getServerSnapshot should be cached" error.
   const serverSnapshotRef = React.useRef<ObservableStatus<T> | undefined>(undefined);
   const getServerSnapshot = React.useCallback<() => ObservableStatus<T>>(() => {
     if (serverSnapshotRef.current === undefined) {
--- a/src/useObservable.ts
+++ b/src/useObservable.ts
@@ -107,7 +107,8 @@ export function useObservable<T = unknown>(observableId: string, source: Observa
   // Reads only `config`, never `observable.immutableStatus`: `preloadedObservables` is a
   // `globalThis` cache keyed only by `observableId`, so a server shares it across concurrent
   // requests, and seeding from it would render one request's data into another's HTML.
-  // The `as` cast below hides a missing `firstValuePromise` from `tsc` and the tests.
+  // React 18 and up only: below that the shim's server path ignores this function and returns
+  // `getSnapshot()`, so the cached value still reaches the markup there.
   // Held in a ref because React requires a stable value across renders.
   const serverSnapshotRef = React.useRef<ObservableStatus<T> | undefined>(undefined);
   const getServerSnapshot = React.useCallback<() => ObservableStatus<T>>(() => {
@@ -121,7 +122,7 @@ export function useObservable<T = unknown>(observableId: string, source: Observa
         data: initialDataValue,
         error: undefined,
         firstValuePromise: observable.firstEmission
-      } as ObservableStatus<T>;
+      };
     }
 
     return serverSnapshotRef.current;
--- a/src/useObservable.ts
+++ b/src/useObservable.ts
@@ -104,7 +104,44 @@ export function useObservable<T = unknown>(observableId: string, source: Observa
     return observable.immutableStatus;
   }, [observable]);
 
-  const update = useSyncExternalStore(subscribe, getSnapshot);
+  // `useSyncExternalStore` requires a third argument when the tree is rendered on a
+  // server or hydrated; without it React throws "Missing getServerSnapshot, which is
+  // required for server-rendered content" and the surrounding subtree falls back to
+  // client rendering.
+  //
+  // This deliberately does NOT return `observable.immutableStatus` the way `getSnapshot`
+  // does. `preloadedObservables` is a `globalThis` cache keyed only by `observableId`, so
+  // on a server it is shared by every concurrent request. Seeding the server snapshot from
+  // it would let one request render data another request fetched for the same path. Only
+  // `config` is safe to read here, because it comes from the caller on this render.
+  //
+  // The result is memoized per component instance because React compares the value it
+  // returns across renders, and a fresh object each time is what triggers the
+  // "The result of getSnapshot should be cached" error.
+  const serverSnapshotRef = React.useRef<ObservableStatus<T> | undefined>(undefined);
+  const getServerSnapshot = React.useCallback<() => ObservableStatus<T>>(() => {
+    if (serverSnapshotRef.current === undefined) {
+      const initialDataValue = config?.initialData ?? config?.startWithValue;
+
+      serverSnapshotRef.current = {
+        status: hasInitialData ? 'success' : 'loading',
+        hasEmitted: hasInitialData,
+        isComplete: false,
+        data: initialDataValue,
+        error: undefined,
+        firstValuePromise: observable.firstEmission
+      } as ObservableStatus<T>;
+    }
+
+    return serverSnapshotRef.current;
+    // `config.initialData` and `config.startWithValue` are read above but deliberately left
+    // out of the dependency array. Callers routinely pass a fresh `config` literal on every
+    // render, so including them would rebuild this callback constantly, and the ref means
+    // the value is computed once per component instance regardless.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [observable, hasInitialData]);
+
+  const update = useSyncExternalStore(subscribe, getSnapshot, getServerSnapshot);
 
   // Return a new object with initialData overlaid rather than mutating the shared
   // _immutableStatus reference, which is the same object across all components
--- a/src/useObservable.ts
+++ b/src/useObservable.ts
@@ -104,31 +104,11 @@ export function useObservable<T = unknown>(observableId: string, source: Observa
     return observable.immutableStatus;
   }, [observable]);
 
-  // `useSyncExternalStore` requires a third argument when the tree is rendered on a
-  // server or hydrated; without it React throws "Missing getServerSnapshot, which is
-  // required for server-rendered content" and the surrounding subtree falls back to
-  // client rendering.
-  //
-  // This applies on React 18 and up. Below that, `use-sync-external-store/shim` ignores the
-  // third argument on the server as well as the client, so both shapes render identically
-  // and nothing here regresses. It also does nothing for it.
-  //
-  // This deliberately does NOT return `observable.immutableStatus` the way `getSnapshot`
-  // does. `preloadedObservables` is a `globalThis` cache keyed only by `observableId`, so
-  // on a server it is shared by every concurrent request. Seeding the server snapshot from
-  // it would let one request render data another request fetched for the same path.
-  //
-  // So no field below carries data across requests: `status`, `hasEmitted` and `data` come
-  // from `config`, which is the caller's own input on this render. `firstValuePromise` does
-  // read the shared `observable`, but it is a `Promise<void>` that resolves without a value,
-  // so it discloses nothing. It is not optional on `ObservableStatus`, and the
-  // `as ObservableStatus<T>` cast below means neither `tsc` nor the tests would notice if it
-  // were dropped: a caller doing `status.firstValuePromise.then(...)` on the server would
-  // just throw.
-  //
-  // The result is memoized per component instance because React compares the value it
-  // returns across renders, and a fresh object each time is what triggers the
-  // "The result of getServerSnapshot should be cached" error.
+  // Reads only `config`, never `observable.immutableStatus`: `preloadedObservables` is a
+  // `globalThis` cache keyed only by `observableId`, so a server shares it across concurrent
+  // requests, and seeding from it would render one request's data into another's HTML.
+  // The `as` cast below hides a missing `firstValuePromise` from `tsc` and the tests.
+  // Held in a ref because React requires a stable value across renders.
   const serverSnapshotRef = React.useRef<ObservableStatus<T> | undefined>(undefined);
   const getServerSnapshot = React.useCallback<() => ObservableStatus<T>>(() => {
     if (serverSnapshotRef.current === undefined) {
@@ -145,10 +125,8 @@ export function useObservable<T = unknown>(observableId: string, source: Observa
     }
 
     return serverSnapshotRef.current;
-    // `config.initialData` and `config.startWithValue` are read above but deliberately left
-    // out of the dependency array. Callers routinely pass a fresh `config` literal on every
-    // render, so including them would rebuild this callback constantly, and the ref means
-    // the value is computed once per component instance regardless.
+    // Callers pass a fresh `config` literal each render, so the fields read above are kept
+    // out of the deps; the ref computes the value once per instance anyway.
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [observable, hasInitialData]);
 
--- a/test/useObservable.test.tsx
+++ b/test/useObservable.test.tsx
@@ -366,8 +366,8 @@ describe('useObservable', () => {
         sink.on('error', reject);
 
         const stream = renderToPipeableStream(<Probe observableId="ssr-streaming" observable$={observable$} />, {
-          onError(e) {
-            error = e;
+          onError(caughtError) {
+            error = caughtError;
           },
           onAllReady() {
             stream.pipe(sink);
--- a/test/useObservable.test.tsx
+++ b/test/useObservable.test.tsx
@@ -446,5 +446,23 @@ describe('useObservable', () => {
       expect(html).not.toContain('first-request-secret');
       expect(html).toContain('loading:undefined');
     });
+
+    it('prefers the callers initialData over a value already in the shared cache', async () => {
+      // The branch above that reads `initialData` is only reachable when the cache ALREADY
+      // holds a value for this id: otherwise the overlay in `useObservable` sets status,
+      // data and hasEmitted itself and the server snapshot never decides anything. So seed
+      // the cache first, exactly as the leak test does, and only then pass `initialData`.
+      const observable$: Subject<any> = new Subject();
+      const observableId = 'ssr-initial-data-beats-cache';
+
+      const { result } = renderHook(() => useObservable(observableId, observable$, { suspense: false }));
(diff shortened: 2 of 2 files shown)


armando-navarro (approved):
Thanks Tyler. This is a good fix and the call you flagged is the right one. I reproduced the failure on `v5` first, then went after the reasoning rather than taking it on trust. Approving. Everything below is non-blocking.

**On the decision you asked about:** you were right to refuse what #748 proposed, and it is worth stating more strongly than "unsafe". Take your fix out, so `getServerSnapshot` returns `observable.immutableStatus` the way the issue asked, and the server HTML comes back as `<div>success:first-request-secret</div>`, a value an earlier render had left in the shared cache. One render's data really does reach another render's markup. Yours does not, which is the point.

### Two things the code gets right that nothing holds in place

- **The `initialData` branch of `getServerSnapshot` is uncovered.** Change it so it always returns `loading`, `false` and `undefined`, ignoring `initialData` entirely, and all 22 tests still pass. "reports initialData on the server" never reaches it, because the overlay at `src/useObservable.ts:149-164` sets `status`, `data` and `hasEmitted` itself whenever `!observable.hasValue && hasData`. Same reason it survived your mutation B. The branch matters when the shared cache already holds a value for that id and the caller also passes `initialData`: the overlay is skipped and your server snapshot is what ships, rendering `success:my-own-data`, correctly preferring the caller's value over the stored one.
- **The natural test is your leak test with a seed added.** Do the first render exactly as `:432` already does, so the shared cache is holding a value for that id, then server-render the same `observableId` with `initialData`, asserting the caller's seed renders and the stored value does not appear. That is the scenario the branch protects, and it doubles as a second cross-request check. I checked that such a test would actually catch a regression rather than just pass: with that branch changed to ignore `initialData`, the same case renders `loading:undefined` and the assertion fails.
- **The ref is uncovered too, though I am flagging that rather than asking for it.** All 22 still pass if the ref is taken out and a fresh object is built on every call, since `renderToString` calls `getServerSnapshot` once and only hydration exercises it. A test for it would have to render on the server, hydrate that markup on the client, and watch what React logs. Your file already spies on `console.error` at `:176`, so the genuinely new part is the hydrating, but that is still more setup than the rest of the suite needs.

### On the ref, the half of your question I nearly skipped

You wrote that the ref keeps the value stable for React's cached-result check. That is right, and it earns its place:

- Take the ref out so a fresh object is built on every call, and hydrating the no-`initialData` case warns `The result of getServerSnapshot should be cached to avoid an infinite loop`. Put the ref back and the same run is silent.
- Small correction: React's string names `getServerSnapshot`, not `getSnapshot` as your description has it.
- Your `exhaustive-deps` suppression looks safe. I tried to make a stale value visible and could not: re-rendering one instance with `initialData` changed renders the new value, because the overlay recomputes from the current `config`. It is the overlay, not the ref, keeping it fresh.

### What I verified

- The failure reproduces on `v5` with your exact error, and **also under streaming, which is what App Router actually uses.** My other checks used `renderToString`, so I redid these on `renderToPipeableStream`: without the third argument `onShellError` fires with the same message and the server emits an empty string, and with your fix the stream completes with zero errors while the case where the shared cache already holds a value still renders `loading:undefined`, with that value absent from the markup.
- Both your mutations reproduce exactly as you reported them.
- Hydration is clean in three cases with zero React warnings, including one where the server said `loading` but the client's cache already held a value. **Your description undersells this:** taking the rendered `status` and `data` from `config` rather than from the cache also makes hydration deterministic, since both sides compute them from the same props.
- Public type surface unchanged: byte-identical `.d.ts` between `v5` and the branch.
- `tsc` clean on both configs, `vite build` fine, eslint 0 errors. `src/useObservable.ts` carries the same 6 warnings as base, so the new source adds none. The test file goes 11 to 16, all of them the existing `no-explicit-any` rule on the new `Subject<any>` annotations.
- Your shim note and your `Fixes #748` note are both correct, so nothing needed on the latter.

### Three smaller notes

- **The comment overclaims a little.** `src/useObservable.ts:115-116` says only `config` is safe to read, but `:132` just below reads `firstValuePromise: observable.firstEmission` from the shared cache. Nothing leaks through it, since that is a `Promise<void>`, and it sits inside the boundary you already drew. Worth changing the wording rather than the code, though. I tried removing the value, and because the field is not optional, `status.firstValuePromise.then(...)` on the server then fails with `Cannot read properties of undefined`. Neither `tsc` nor the suite catches that, because the `as ObservableStatus<T>` cast covers the missing field.
- **The protection is React 18 and up.** Below that the shim ignores the third argument on the server too, not only on the client: `useSyncExternalStore$1` names it and then returns `getSnapshot()`. On React 17 both forms render identically so nothing regresses, but a clause scoping the comment to 18+ would stop the next reader over-trusting it. I had this backwards at first.
- **Suspense SSR is still unfixed, and without `initialData` server rendering produces only `loading` placeholders.** With suspense on, the throw at `:80` happens before the hook, so `getServerSnapshot` never runs: cold cache throws `A component suspended while responding to synchronous input`, with `initialData` it renders, and suspense off renders. Both are the right scope, and the placeholder behavior is the necessary cost of your model rather than a flaw in it, since `observableId` alone cannot tell your own data from another request's. A line in the description would stop anyone reading the title as "SSR works now".

Approving. If I have misread anything, especially the two coverage gaps, point me at it and I will look again.

tyler-reitz:
You read all of it right, and the coverage gap was real. Pushed `f2882a3` on top of your approval, so here is what changed.

**I reproduced the gap before acting on it.** Neutering the branch to always return `loading` / `false` / `undefined` leaves all 22 passing, exactly as you said. Your proposed test is in, and I wrote it against the neutered branch first so a pass would mean something: it fails there with `loading:undefined`, and passes against the real branch. 23 now.

Your diagnosis of why was the useful part. The overlay handles the ordinary case, so the branch only decides anything when the cache already holds a value for the id **and** the caller passes `initialData`. I would not have found that from the failing direction.

**The comment is fixed on both counts.** It now says why `firstValuePromise` is safe rather than claiming only `config` is read, and it records your point that the `as ObservableStatus<T>` cast would hide that field going missing from both `tsc` and the suite. `getServerSnapshot` name corrected. I added your React 18+ clause too.

**The description now carries the scope caveat**, in its own section: suspense mode still cannot server-render, and without `initialData` a cold cache renders `loading` placeholders. Stated as inherent to the model rather than a gap, for the reason you gave.

**Not done: the ref coverage.** You flagged it rather than asked, and I agree with your reasoning about the setup cost, so I have left it. Worth revisiting when something else in the suite needs a hydration harness.

Two things of yours I am carrying forward rather than answering here. The `renderToPipeableStream` check is the more valuable verification, since App Router is the real target and my own checks were all `renderToString`. And your point that taking `status` and `data` from `config` makes hydration deterministic is a better argument for this shape than the one I wrote, so I have taken it.

CI has not run on this push. GitHub Actions has been in a major outage since 15:22 UTC and no workflow runs are being created, so the checks you see are stale against `9f26b11`. Locally: both typechecks clean, 23/23.

tyler-reitz:
Your `renderToPipeableStream` check sent me back to re-run some unrelated SSR work of mine that was verified on `renderToString` only. It found something that belongs on this PR, because it is downstream of your review and it is about this code.

**In suspense mode under streaming, this hook waits for the data and then renders the placeholder anyway.**

The sequence: the component suspends at `useObservable.ts:80`, React holds the boundary open, the observable emits, the thrown `firstEmission` resolves, React retries. **The retry is still a server render**, so `useSyncExternalStore` reads `getServerSnapshot`, which this PR deliberately keeps away from the shared cache. The value that just arrived is discarded and the boundary resolves to `loading:undefined`. React's `<!--$-->` resolved-boundary marker sits right next to it in the markup, so the suspend really did resolve rather than fall back.

Net effect: the caller pays the full latency and gets the placeholder.

**This is not an argument against the PR and I am not proposing a change here.** Reading that cache is the cross-request leak you reproduced at the top of your review. Refusing to read it is right, and the alternative is the `first-request-secret` markup you demonstrated. What it means is that per-request cache scoping is the thing that would let the server render real data, rather than a separate correctness cleanup, since with a request-scoped cache `getServerSnapshot` could safely read the request's own entry.

One other thing from the same run, since it is worse operationally than what we knew: **with a cold cache in suspense mode, streaming HANGS rather than erroring.** Under `renderToString` you get `A component suspended while responding to synchronous input`. Streaming treats suspending as legitimate and holds the response open, and `firstEmission` only resolves on a first emission, so an observable that never emits never completes the request. A hung request rather than a crash, and no test we have could have seen it.

Both of these sharpen the scope note I added to the description rather than contradicting it. Nothing here needs action from you, and the approved diff is unchanged.

Also worth saying plainly: my SSR checks were all `renderToString`, and yours were not, which is why you found things I did not. I have written that down as a rule rather than a one-off.

test/useObservable.test.tsx
```diff
@@ -391,4 +392,77 @@ describe('useObservable', () => {
...
+    };
+
+    it('renders on the server instead of throwing', () => {
+      const observable$: Subject<any> = new Subject();
+
+      // Without a getServerSnapshot, React throws "Missing getServerSnapshot, which is
+      // required for server-rendered content" and the whole subtree falls back to client
+      // rendering. This is the #748 regression test: delete the third argument to
+      // useSyncExternalStore and this assertion fails.
+      expect(() => renderToString(<Probe observableId="ssr-renders" observable$={observable$} />)).not.toThrow();
```
jhuleatt, line 348: can you confirm that this test fails in the current version of reactfire? just want to make sure `renderToString` does SSR the same way as a Server Component would

src/useObservable.ts
```diff
@@ -104,7 +104,55 @@ export function useObservable<T = unknown>(observableId: string, source: Observa
     return observable.immutableStatus;
   }, [observable]);
 
-  const update = useSyncExternalStore(subscribe, getSnapshot);
+  // `useSyncExternalStore` requires a third argument when the tree is rendered on a
```
jhuleatt: please clean up/shorten comments to just the critical info a future reader of the code would need

jhuleatt (commented):
I think this can target the main branch as a v4 bugfix instead of v5. Sound ok to you?

tyler-reitz:
Heads up: the approval above predates this diff. Retargeted from `v5` to `main` (Jeff called it a bug fix rather than a breaking change) and force-pushed.

Rebased onto `main`, which drops the `v5`-only test imports, and pruned the comments in `useObservable.ts` from 30 added lines to 8. The `as ObservableStatus<T>` note you raised is kept as a one-liner in the source and is still unresolved.

The body's counts were wrong once `v5`'s tests left the file, and have since moved again with a streaming test added at `22feb8e`. All re-run on this base: six tests in the block, all six fail on dropping the third argument, and returning `immutableStatus` fails two of them. Suite is 20/20.

Not asking for a re-review, but happy to have one.

src/useObservable.ts
tyler-reitz: Done in `dc9884e`, 30 added comment lines down to 8:

```ts
// Reads only `config`, never `observable.immutableStatus`: `preloadedObservables` is a
// `globalThis` cache keyed only by `observableId`, so a server shares it across concurrent
// requests, and seeding from it would render one request's data into another's HTML.
// The `as` cast below hides a missing `firstValuePromise` from `tsc` and the tests.
// Held in a ref because React requires a stable value across renders.
```

The first three lines are the ones I'd argue to keep: returning `observable.immutableStatus` is the obvious implementation and it leaks across requests, so without that note it reads like a needless detour. The `as` cast line is Armando's open review note.

test/useObservable.test.tsx
tyler-reitz: Confirmed both halves.

Fails on current reactfire: reverted `src/useObservable.ts` to `upstream/main`, which has no `getServerSnapshot` in it, and ran the new block against that. All six fail with `Missing getServerSnapshot`.

On the renderer, good instinct, so I added a test rather than argue it. `22feb8e` covers `renderToPipeableStream` alongside `renderToString`, using a real `Writable` sink:

| Renderer | current `main` | this PR |
| --- | --- | --- |
| `renderToString` | fails | passes |
| `renderToPipeableStream` | fails | passes |

Same error both ways, so `renderToString` wasn't flattering it. Test-only, about 30 lines.

armando-navarro (approved):
One thing I would fix before merge, and it is a description edit, not code.

### The description's suspense scope section

The bolded paragraph says suspense mode is unchanged, that the throw at `src/useObservable.ts:80` happens before the hook runs so `getServerSnapshot` is never reached, and that only `initialData` gets you a render. I ran the cases and each of those three is contradicted:

- The suspend decision is `hasData = observable.hasValue || hasInitialData`. Whenever the shared cache already holds a value for the id, the throw is skipped, the hook runs, and `getServerSnapshot` is reached in suspense mode.

- On `main`, that already-cached case throws `Missing getServerSnapshot`. On this branch it renders `loading:undefined`. So suspense mode's server behavior did change: a named crash became a silent placeholder, with no `initialData` involved.

- Your own August 6 comment describes the streaming version of the same thing: suspend, wait for the emission, retry reads `getServerSnapshot`, placeholder ships anyway. The description asserts the opposite of what that comment measured.

- Since a suspense-mode cache entry never expires, and even a first render that crashes has already seeded the entry (the `SuspenseSubject` constructor subscribes on its own), the already-cached case is the steady state on a long-lived server process, not an edge.

- One more sentence in the same paragraph is narrower than it reads: a cold cache "fails with `A component suspended...`" under `renderToString` only. Your August 6 comment already records that under streaming the same case hangs instead, so the corrected paragraph is worth writing against both renderers.

The cost worth writing down is diagnosability: a named library error becomes silent loading HTML, or a `TypeError` in the consumer's component when it dereferences `data` (suspense mode's contract is that the hook does not return until data exists, so callers dereference it). The description just needs to say what suspense mode now does on the server, rather than that this path cannot run.

### Optional, in rough order of value

- **The `as ObservableStatus<T>` cast on the server snapshot literal can come out, and taking it out retires your open note.** Removing it compiles clean on both tsconfigs (`initialData` is typed `T | any`, so `data` already fits), and once it is gone, deleting `firstValuePromise` fails `tsc` with TS2741, so the compiler takes over what your comment line documents by hand and the line can go with the cast.

- **The pruned comment reads as an unconditional guarantee, but it holds only where React calls `getServerSnapshot`, which is React 18+.** On 16 and 17 (the peer range allows `>=16`) the shim's server path literally `return getSnapshot()`, so the served value comes from the shared cache and the cross-request behavior the comment rules out is still reachable there. That is unchanged from `main` and not this PR's job to fix, but it is a real reachable path, so the scoping clause your `6e42acf` message added (and the prune removed) might be worth restoring.

- **If you edit the squash message at merge:** `27dfc44`'s body still describes the block it ships as "four tests", with the `immutableStatus` version passing three and failing only the leak test. The block is now six tests, and that mutation fails two.

- **`onError(e)` in the streaming test:** the enclosing test already declares `error`, so a fuller name like `caughtError` would read better than the single letter.

- **Hydration coverage remains the open gap.** On this base the suite still stays green with the ref made inert, matching round 1's result. React only warns on an uncached server snapshot rather than failing, so a `hydrateRoot` test would need to assert on `console.error`, which is the setup cost I called in round 1 and you agreed. Still your call.

Everything else I checked matches what you posted. If I have misread the suspense cases, point me at the case and I will rerun it.

tyler-reitz:
You are right on all three, and the correction is in. I ran your cases on both branches rather than reasoning about them, with `main` as the control:

| `suspense: true` | `main` | this branch |
| --- | --- | --- |
| Cold cache, `renderToString` | suspended-while-sync throw | same |
| Cold cache, streaming | hangs | same |
| Value cached for the id, `renderToString` | `Missing getServerSnapshot` | `loading:undefined` |
| Value cached for the id, streaming | `onShellError`, empty output | completes with `loading:undefined` |
| Value cached, consumer dereferences `data` | `Missing getServerSnapshot` | `TypeError: Cannot read properties of undefined` |

Your steady-state point is the one I had backwards: seeding the entry does not need a successful render, since the `SuspenseSubject` constructor subscribes even when the render throws. The description now says what suspense mode does on the server, with the diagnosability cost named.

**Cast removed**, and it retires the note: TS2741 now fires if `firstValuePromise` goes missing, so the comment line documenting that by hand is gone too. **The React 18+ clause is restored** in the source comment. **`caughtError`** renamed. I will fix the "four tests" count in the squash message at merge.

Hydration coverage still not taken, same reasoning as round 1.

Suite 20/20, both mutations re-run on this diff: dropping the third argument fails all six, returning `immutableStatus` fails two.

PR #780, opened 2026-08-05
ci: add a manual firestore flake probe for #776

tyler-reitz:
Refs #776. This measures the flake; it does not fix it.

## Why

Every measurement of the #776 flake so far has been on a laptop. Locally the failure is a plain `waitFor` timeout with no gRPC error at all. In CI it arrives alongside a gRPC framing desync:

```
RESOURCE_EXHAUSTED: Received message larger than max (795107700 vs 4194304)
```

A ~795 MB claimed frame against a 4 MB cap means the reader is interpreting payload bytes as a length header. That is a `@grpc/grpc-js`-layer failure, and it does not appear in any local run.

**So the local repro and the CI failure may not be the same bug.** That matters, because the `@grpc/grpc-js` override proposed as the fix was measured only against the local one, at p = 0.12. Merging that override would also mean CI never runs 1.9.16 again, so it is a before-and-after with no control.

This workflow makes the comparison in CI, where the failure actually happens, before anything is changed.

## What it does

Matrixes Node 22 and 24. Within each job it runs **both `@grpc/grpc-js` arms sequentially**, baseline then override, N times per arm (default 30), and writes a table per arm to the job summary: resolved grpc-js version, pass / flake / hang / infra counts, how many runs showed `RESOURCE_EXHAUSTED`, how many flakes showed it too, and the resulting rate.

Design decisions worth reviewing:

- **`workflow_dispatch` only.** It never runs on a push, a PR, a release or a schedule, so it adds nothing to normal CI time.
- **The arms are deliberately NOT a matrix dimension.** The premise of the probe is that the machine matters, so putting the arms on separate runners would reintroduce the variable being measured. They run back to back on one runner instead, which costs a longer job. Arm order is forced baseline-then-override, because applying the override mutates `node_modules` for everything after it.
- **A fresh emulator per iteration**, matching how `npm test` runs in CI. Reusing one emulator across iterations would measure something else.
- **Failures are classified, not counted.** Only the #776 assertion signature counts toward the rate. Emulator start failures are reported separately, because folding them in is exactly what inflated the earlier local estimate by roughly 50% and made 30 runs look adequately powered when they were not.
- **Hangs get their own bucket.** A test timeout produces no assertion line, so without this it would land in `infra` and drop out of the rate entirely. #776 reports a 120s hang.
- **Unrecognized failures are surfaced, not swallowed.** The flake match is a literal vitest assertion string, so a reword would quietly turn every real flake into an infra failure. Anything unmatched has its assertion line collected and shown as a warning in the summary.
- **It reports rather than fails.** A red job here means the probe broke, not that the flake reproduced. It exits non-zero only if every run failed for infrastructure reasons, meaning it measured nothing.
- **The summary renders in an `always()` step** from counts persisted per arm, so a job killed by `timeout-minutes` still reports the arms that finished instead of losing every count.
- **`RESOURCE_EXHAUSTED` is counted per arm, and its overlap with the flake count is counted too.** Two independent totals cannot say whether the gRPC desync and the #776 assertion co-occur, which is the question the probe exists for.
- Inputs reach the script through `env` rather than being interpolated into the shell, and both `iterations` and `arms` are validated before they reach the loop.

⚠️ **This probes one emulator and one test file. CI runs five emulators and the whole suite in parallel.** `RESOURCE_EXHAUSTED` has never appeared in a firestore-only run, so this configuration may reproduce the local failure while never reaching the CI one. Every job summary repeats this, because a clean table here is not a verdict on CI.

## Verification

⚠️ **This workflow cannot be exercised from this PR.** GitHub only offers `workflow_dispatch` for workflows present on the default branch, so it is unrunnable until this merges. Per the repo's own lesson about treating jobs authored on a branch as unrun, it was verified out of band instead:

- **The classifier was dry-run against synthetic logs** covering pass, flake, flake-with-gRPC-error, hang, infra-failure and a *reworded* assertion. It returned the expected counts, excluded hangs and infra failures from the rate, and routed the reworded assertion to the unmatched warning rather than counting it. This is the part most worth getting right, since a probe that miscounts is worse than no probe.
- **The summary was rendered from a partial counts file**, which is the path a job killed by `timeout-minutes` takes.
- **The `arms` input rejects** empty, unknown and non-JSON values, and normalizes to baseline-first regardless of the order given.
- **zizmor 1.25.2** reports no findings beyond the `cache-poisoning` rule CI suppresses, which is the severity set the gate reads.
- YAML parses; the `package.json` override edit produces the right key (`npm pkg set` mangles keys containing a slash, hence the `node -e`); `npx firebase` resolves through `firebase-tools`.

**The first real run is therefore the first proof it works.** If it misbehaves, the fix is another PR, not a rollback of anything consumers see.

## Known confound: narrower than first stated

**Corrected after review.** This section originally warned that the repo-wide override moves `@grpc/grpc-js` for `firebase-tools` and `google-gax` as well as the Firestore client, so the arms differed by more than the library under test. A lockfile diff of the two arms shows that is not the case: **exactly one version moves, `@grpc/grpc-js` 1.9.16 to 1.14.4.** `google-gax` was already on 1.14.4, so its nested copy simply dedupes into the hoisted one. Nothing else budges.

So the arms differ by the Firestore client's grpc-js and nothing else, which is a cleaner comparison than the original caveat implied. Credit to @armando-navarro for diffing it rather than taking the caveat at face value.

## What this changes for anyone else

Nothing. One new file, no existing workflow touched, no runtime or packaging change, and no effect on normal CI.

### DIFF
--- a/.github/workflows/flake-probe.yaml
+++ b/.github/workflows/flake-probe.yaml
@@ -6,6 +6,10 @@
 # not the same bug. This runs both `@grpc/grpc-js` arms on both Node versions under CI
 # conditions so the comparison is made where the failure actually happens.
 #
+# Both arms run inside a single job, sequentially on the same runner. They are
+# deliberately NOT a matrix dimension: the whole premise is that the machine matters,
+# so splitting the arms across two runners would reintroduce the variable being tested.
+#
 # Manual only. It never runs on a push, a PR or a schedule, so it costs nothing until
 # someone asks for it.
 name: Firestore flake probe
@@ -16,7 +20,7 @@ on:
       iterations:
         description: "Test runs per arm (each is a full emulator start/stop, roughly 25s)"
         required: false
-        default: "20"
+        default: "30"
       node_versions:
         description: "JSON array of Node majors to probe"
         required: false
@@ -37,9 +41,8 @@ jobs:
     strategy:
       matrix:
         node: ${{ fromJSON(inputs.node_versions) }}
-        arm: ${{ fromJSON(inputs.arms) }}
       fail-fast: false
-    name: Probe Node ${{ matrix.node }} / ${{ matrix.arm }}
+    name: Probe Node ${{ matrix.node }}
     steps:
       - name: Checkout
         uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
@@ -68,32 +71,16 @@ jobs:
       - name: Install deps
         run: npm ci
 
-      # `npm pkg set` mangles keys containing a slash, so edit package.json directly.
-      # `npm install` (not `npm ci`) is required here because applying an override
-      # necessarily changes the lockfile.
-      - name: Apply the grpc-js override
-        if: ${{ matrix.arm == 'override' }}
-        run: |
-          node -e '
-            const fs = require("fs");
-            const pkg = JSON.parse(fs.readFileSync("package.json", "utf8"));
-            pkg.overrides = { ...pkg.overrides, "@grpc/grpc-js": "^1.14.0" };
-            fs.writeFileSync("package.json", JSON.stringify(pkg, null, 2) + "\n");
-          '
-          npm install --no-audit --no-fund
-
-      - name: Record the resolved grpc-js version
-        run: |
-          RESOLVED="$(node -p "require('@grpc/grpc-js/package.json').version")"
-          echo "resolved_grpc=$RESOLVED" >> "$GITHUB_ENV"
-          echo "Resolved @grpc/grpc-js: $RESOLVED"
-
       # Inputs and matrix values are passed through `env` rather than interpolated into
       # the script body, so nothing from the dispatch form can be executed as shell.
+      #
+      # Counts are appended to `probe-counts.tsv` as each arm finishes, and the summary
+      # is rendered by a separate `always()` step. If the job hits its timeout partway
+      # through, whatever was measured before the wall still gets reported.
       - name: Run the probe
         env:
           ITERATIONS: ${{ inputs.iterations }}
-          ARM: ${{ matrix.arm }}
+          ARMS: ${{ inputs.arms }}
           NODE_MAJOR: ${{ matrix.node }}
         run: |
           set -uo pipefail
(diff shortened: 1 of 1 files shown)


armando-navarro (approved):
Nice, this is worth having. Approving it. A few things I would think about before anyone runs it and reads numbers off the table.

### The two arms land on different runners

`arm` is a matrix dimension, so baseline and override come out as separate jobs on separate machines. You were careful about this on your laptop, same machine and back to back, and that is the one control that did not make the trip. Given the whole premise here is that the machine matters, I would pull `arm` out of the matrix and loop both arms inside one job instead. Costs you a longer job, but then the comparison means something.

### It is not running the workload the failures came from

CI does `npm run test`, so five emulators, the whole suite, parallel workers. This does one emulator and one file.

Your local runs were firestore-only too and still hit 6/44, so I am not saying it will fail to reproduce anything. But `RESOURCE_EXHAUSTED` has never turned up in a firestore-only run, and that is the signal you actually care about. What worries me is that you reproduce the laptop bug on CI hardware and the CI bug never shows up. Might be worth a line in the job summary so nobody reads a clean table as a verdict.

### The RESOURCE_EXHAUSTED number is a total, not an overlap

`grpc_err` goes up before the run gets classified, so `Flake: 2` sitting next to `RESOURCE_EXHAUSTED: 11` does not tell you whether those 2 are inside the 11. That overlap is the whole question. One more counter in the flake branch and it is in the table.

### Hangs fall out of the rate entirely

If Firestore calls never finish you get `Error: Test timed out in 5000ms` and no assertion line, so the run goes to `infra` and drops out of both halves of the rate. #776 mentions a 120s hang. Should those get their own bucket?

Related: the flake match is a literal assertion string, so if vitest ever rewords it, every real flake quietly becomes an infra failure.

### On run counts

#776 calls 30 per arm underpowered at p = 0.12, and the default here is 20.

Separately, the guard lets you pass 200. At your 25s estimate that is 83 minutes against a 60 minute timeout. I clocked 7s locally so your estimate is probably pessimistic, but if it ever does hit the wall the summary block never runs and you lose every count.

### Your confound is not actually there

I diffed the two lockfiles and exactly one version moves: `@grpc/grpc-js`, 1.9.16 to 1.14.4. `google-gax` was already on 1.14.4, so its nested copy just dedupes into the hoisted one. Nothing else budges. The arms differ by the Firestore client's grpc-js and nothing else, which is cleaner than your caveat makes it sound.

One last thing. I checked the classifier by actually breaking the test rather than trusting the grep on sight. Shortened the `waitFor` on `returns undefined if document does not exist` and ran your command against it. Your pattern catches it.

tyler-reitz:
Thanks, all six are in as `43f39e5`. Taking them in order.

**The two arms landing on different runners.** You're right and this is the one that mattered most: the premise is that the machine matters, so splitting the arms across runners reintroduces the variable being measured. `arm` is out of the matrix; both arms now run sequentially inside one job. Matrix is Node alone. Arm order is forced baseline-then-override regardless of what the input says, because applying the override mutates `node_modules` for everything after it, and I added a header comment saying why the arms are not a matrix dimension so it doesn't get put back.

**Not running the workload the failures came from.** Kept the scope, took the line in the summary. Every run now ends with a footer saying this is one emulator and one file against CI's five emulators and parallel suite, that `RESOURCE_EXHAUSTED` has never appeared in a firestore-only run, and that this configuration may reproduce the local failure while never reaching the CI one. A clean table is explicitly not a verdict.

**The RESOURCE_EXHAUSTED total vs the overlap.** Added a `flake_with_grpc` counter in the flake branch, rendered directly under the flake row as "...of which also showed RESOURCE_EXHAUSTED". Two independent totals could not answer the question the probe exists for.

**Hangs falling out of the rate.** They get their own bucket now, matching `Test timed out in Nms` and `Hook timed out in Nms`, reported and excluded rather than silently absorbed into `infra`. Your related point about the literal assertion string is the sharper one, so it's handled too: any failure the classifier does not recognize gets its assertion line captured and rendered as a warning block in the summary. If vitest rewords the message, it shows up as "the flake counts above are wrong" instead of quietly inflating `infra`. I verified that path with a synthetic log carrying `strictly equal` instead of `deeply equal`, and it lands in the warning rather than being counted.

**Run counts.** Default is 30 now. On the 200 guard: rather than lower the cap, counts are appended per arm to a `probe-counts.tsv` and the summary renders in a separate `always()` step, so a job killed at the wall still reports every arm that finished instead of losing everything. Tested against a partial counts file.

**The confound.** Thanks for diffing the lockfiles, that is better than the caveat I wrote. I'll narrow the "Known confound" section in the body to what you found: only `@grpc/grpc-js` moves, 1.9.16 to 1.14.4, and `google-gax`'s nested copy dedupes into the hoisted one.

Also dropped a `set +e` / `set -e` pair around the test invocation that was turning errexit on partway through a script that never had it enabled. Latent, never fired, but it went with the restructure.

Re-verified after the changes rather than assuming: classifier dry-run over synthetic logs covering pass, flake, flake-with-gRPC, hang, infra and the reworded assertion returns the expected counts; the arms input rejects empty, unknown and non-JSON values; zizmor 1.25.2 still reports nothing beyond the suppressed `cache-poisoning` rule. The first real run is still the first proof of the whole thing, since `workflow_dispatch` isn't offered until this is on the default branch.

And noted on checking the classifier by actually breaking the test rather than reading the grep. That's the second time this week your verification caught something mine wouldn't have.

Issue #783, opened 2026-08-07
Remove the firestore flake probe workflow once #776 is resolved


tyler-reitz:
`.github/workflows/flake-probe.yaml` was added in #780 as a temporary measurement tool for #776. It is `workflow_dispatch` only, so it costs nothing until someone runs it, but it is not meant to live in the repo indefinitely. This issue is its removal ticket, filed at merge time so the intent does not survive only in a PR body.

## Remove it when

**#776 is resolved**, not merely measured. Measuring the flake rate is not fixing it: between the measurement and the fix sits the actual `@grpc/grpc-js` override change, and the probe is the thing that verifies that change landed. So it outlives the first run by design.

Concretely, it can go once all of these hold:

- [ ] The probe has run in CI and produced a rate for both arms.
- [ ] A decision has been taken on the `@grpc/grpc-js` override, either way.
- [ ] If the override lands, the probe has been re-run to confirm the rate moved.
- [ ] Removal PR deletes `.github/workflows/flake-probe.yaml`. Nothing else references it.

## Two reasons this is unlikely to be same-day

**1. The default run is underpowered.** #776 already found 30 runs per arm inadequate at p = 0.12, and 30 is the default here. At that rate you would expect roughly 4 flakes in 30 baseline runs against 0 in the override arm, which does not reach significance. Reaching it needs closer to 50 per arm. At an estimated 20-25s per iteration (one emulator boot plus one test file; for calibration, the full five-emulator suite runs in ~57s in CI), 50 per arm is roughly 47 minutes of wall clock against a 60 minute job timeout. Plan on more than one run.

**2. The probe may not be able to see the CI bug at all.** It runs one emulator and one test file; CI runs five emulators and the whole suite in parallel. `RESOURCE_EXHAUSTED: Received message larger than max` has never appeared in a firestore-only run, and that signal is the entire reason to suspect the local repro and the CI failure are different bugs. If the first run comes back with a clean `RESOURCE_EXHAUSTED` column, that is not evidence the CI bug is absent. It is evidence this configuration cannot reach it, and the response is to widen the probe toward the full suite, which is a second change to this workflow rather than a deletion.

The job summary states this caveat on every run so a clean table is not read as a verdict.

## Suggested first run

[...]

PR #785, opened 2026-08-07
ci(flake-probe): add a full-suite workload mode

tyler-reitz:
Refs #776, #783. Adds a `full-suite` workload mode to the flake probe merged in #780.

## Why

The probe ran for the first time today: [run 31204989659](https://github.com/FirebaseExtended/reactfire/actions/runs/31204989659), 30 iterations per arm, both Node versions. **120 of 120 runs passed, with zero `RESOURCE_EXHAUSTED`.**

That is a negative result rather than good news. The probe runs one emulator and one test file; the #776 failures come from `npm run test`, which starts five emulators and runs the whole suite with parallel workers. Zero flakes in 60 baseline runs excludes the 13.6% local rate for this configuration (`0.88^60` is about 0.05%), and by the rule of three the true rate here is under 5%.

Two consequences:

1. **The override arm is uninformative.** It cannot improve on a baseline that never failed, so there is still no controlled evidence either way on the `@grpc/grpc-js` bump.
2. **The isolated suite cannot serve as a control at all**, so merging the override would foreclose the comparison rather than merely complicate it.

The fix is to run the workload the failures actually come from.

## What this changes

- **New `workload` input**, a choice of `full-suite` (the new default) or `firestore-only`. Full-suite runs `npx firebase emulators:exec --project=rxfire-525a3 "npx vitest run"`, which is what `test.yaml` runs. `firestore-only` is kept rather than deleted: it isolates the Firestore client and is roughly 3x cheaper per iteration, which is still useful for a targeted question.
- **Functions deps are installed in full-suite mode.** `test.yaml` does `npm install` in `./functions` before `npm run test`, and the functions emulator does not start without it. The probe was missing this, which would have failed every full-suite iteration.
- **`timeout-minutes` 60 -> 180.** See the arithmetic below.
- **A failing `double check - emulator is running` is now classified as infra**, ahead of the hang check.
- **The summary footer is workload-aware.** A firestore-only table now carries a warning naming the 120/120 result and pointing at full-suite.

## Correcting the timing figure I published

I previously told #783 and #776 that a full-suite iteration costs about 57s and that this collided with the 60 minute timeout. **That was wrong, and both comments have been corrected.** 57s was the duration of the whole `Test Node.js 22` job including checkout, `npm ci`, Java setup and artifact download. The `Run tests` step itself is **23s**.

| Mode | Per iteration | 30/arm (60 runs) | 50/arm (100 runs) | 200/arm (400 runs) |
| --- | --- | --- | --- | --- |
| `firestore-only` | ~8s | ~10 min | ~16 min | ~55 min |
| `full-suite` | ~23s | ~28 min | ~43 min | **~153 min** |

So the default fits comfortably inside the old 60 minute timeout and the raise is not needed for it. **The raise is headroom for the 200 cap**, which is the only setting that can hit the wall. The `always()` summary step added in #780 means a run that does hit it still reports the arms that finished.

## The classification bug, and how it was found

Running one full-suite iteration locally and feeding the real log to the classifier is what surfaced this. It came back **`hang`**, which is wrong: the failure was the RTDB emulator being unreachable, and an emulator that never came up is infra.

It reached `hang` because the health check fails *by timing out*, so it matched `Test timed out in 5000ms`. `test/auth.test.tsx`, `test/firestore.test.tsx` and `test/database.test.tsx` each open with one of these health checks, so in full-suite mode this would have fired regularly and quietly inflated the hang count, which is the bucket meant for #776's 120s hang.

The check now precedes the hang branch and is anchored on `FAIL.*`, so a *passing* health check in the same log does not swallow a real #776 assertion.

**Inspection would not have found this.** It only showed up against a real log.

## Verification

- **One full-suite iteration run locally, end to end.** 8 of 9 test files pass; the failure is the unreachable local RTDB emulator, which is what produced the log above. This also confirms the command works and that the functions deps are required.
- **That real log fed through the classifier**, before and after the fix: `hang` before, `infra (emulator health)` after.
- **Regression cases unchanged**: flake, hang, unrecognized-infra, and a log where the health check passes alongside a genuine #776 assertion, which still classifies as `flake`.
- **The `workload` input rejects unknown values** before reaching the loop, and the two known values produce the expected emulator and vitest arguments.
- **zizmor 1.25.2**: no findings beyond the suppressed `cache-poisoning` rule.

⚠️ **As with #780, the first real full-suite run is the first proof.** `workflow_dispatch` only offers workflows on the default branch, so this cannot be exercised from the PR.

## What this changes for anyone else

Nothing. One workflow file, manual dispatch only, no effect on normal CI.

### DIFF
--- a/.github/workflows/flake-probe.yaml
+++ b/.github/workflows/flake-probe.yaml
@@ -51,13 +51,20 @@ permissions:
 jobs:
   probe:
     runs-on: ubuntu-latest
-    # Measured 2026-08-07: a full-suite iteration is ~23s (the `Run tests` step of the
-    # real CI job, not the whole job, which is ~57s including install and setup), and a
-    # firestore-only iteration is ~8s. The default 30 per arm across two arms is
-    # therefore ~28 minutes full-suite. The 200 cap is what needs the headroom: 400
-    # full-suite iterations is roughly 153 minutes. The `always()` summary step means a
-    # run that does hit this wall still reports the arms that finished.
-    timeout-minutes: 180
+    # Sized 2026-08-07 on the `Run tests` step of the real CI job (not the whole job,
+    # which is ~57s including install and setup).
+    #
+    # ⚠️ Do NOT size this on a single figure. Across recent CI runs that step lands
+    # anywhere from 20 to 31 seconds, and a timeout cares about the slow tail, not the
+    # median. At 31s the 200 cap wants ~207 minutes, which a flat-23s estimate (~153)
+    # would have put comfortably inside 180. And if #776's reported ~120s hang ever
+    # reproduces, those iterations cost far more than any of this.
+    #
+    # The default 30 per arm is ~28-38 minutes full-suite and is safe under any reading.
+    # This ceiling exists for the 200 cap, so it is set past the pessimistic figure
+    # rather than the average one. The `always()` summary step means a run that does hit
+    # the wall still reports the arms that finished.
+    timeout-minutes: 240
     strategy:
       matrix:
         node: ${{ fromJSON(inputs.node_versions) }}
@@ -112,6 +119,12 @@ jobs:
           WORKLOAD: ${{ inputs.workload }}
           NODE_MAJOR: ${{ matrix.node }}
         run: |
+          # ⚠️ errexit is ON here even though nothing below turns it on: GitHub runs an
+          # undeclared `run:` step as `bash -e {0}`, and `set -uo pipefail` does not
+          # disable it. Every command that is ALLOWED to fail therefore has to say so.
+          # Getting this wrong means the step dies on the first failing iteration and
+          # the probe can only ever report a clean table, which is the one failure mode
+          # that makes the whole workflow useless. See the loop below.
           set -uo pipefail
 
           # Guard against a non-numeric or absurd `iterations` before it reaches the loop.
@@ -190,9 +203,18 @@ jobs:
               # Reusing one emulator across iterations would measure a different thing.
               # Unquoted on purpose: both are either empty or a fixed literal set above,
               # never user input.
+              #
+              # ⚠️ `set +e` is LOAD-BEARING, do not remove it. This command failing is the
+              # entire point of the probe, but the step runs under `bash -e`, so without
+              # this the first flake kills the step before `rc` is even read: no
+              # classification, no tally for the arm, and an empty or half-written
+              # probe-counts.tsv. It was removed once on the reasoning that the script
+              # never sets `-e` itself, which is true and irrelevant.
+              set +e
               npx firebase emulators:exec $EMULATOR_ARGS --project=rxfire-525a3 \
                 "npx vitest run $VITEST_ARGS" > "$log" 2>&1
               rc=$?
+              set -e
 
               saw_grpc=0
               if grep -q "RESOURCE_EXHAUSTED: Received message larger than max" "$log"; then
@@ -200,23 +222,41 @@ jobs:
                 saw_grpc=1
               fi
 
+              # ⚠️ EVERY QUESTION BELOW IS SCOPED TO FIRESTORE'S OWN OUTPUT, and it has to
+              # be. In firestore-only mode anything in the log was necessarily about #776.
+              # In full-suite mode that is false: `expected 'loading' to deeply equal
+              # 'success'` is just what vitest prints when a data hook's status assertion
+              # fails, and it appears in 6 of the 9 test files, 40 times over. An unscoped
+              # search counts a slow storage upload or a functions failure as a #776 flake.
+              #
+              # In a vitest log the FAIL line names the file and the assertion or timeout
+              # sits on the NEXT line, verified against #781's real overnight failure, so
+              # -A1 is the correct window. Captured into a variable rather than piped into
+              # `grep -q`, because an early-exiting `grep -q` can SIGPIPE its producer and
+              # `pipefail` would turn that 141 into a silent "no match".
+              fs_fails="$(grep -A1 -E "FAIL.*test/firestore\.test\.tsx" "$log" || true)"
+
               if [ "$rc" -eq 0 ]; then
                 pass=$((pass + 1))
                 echo "run $i: PASS"
-              elif grep -qE "FAIL.*double check - emulator is running" "$log"; then
+              elif grep -qE "FAIL.*test/firestore\.test\.tsx.*double check - emulator is running" "$log"; then
                 # `test/{auth,firestore,database}.test.tsx` each open with an emulator
-                # health check. If one of those fails, the emulator did not come up and
-                # the whole iteration is void, so it is infra rather than a result.
+                # health check. If FIRESTORE's fails, its emulator did not come up and no
+                # firestore result this iteration means anything, so the run is void.
                 #
-                # This has to precede the hang check: a health check that fails does so
-                # by timing out, so without this it would land in `hang` and be read as
-                # the #776 120s hang. Found by running the real suite against a machine
-                # where the RTDB emulator was unreachable.
+                # ⚠️ Scoped to firestore deliberately. An unscoped check let ANY emulator's
+                # health failure outrank a real firestore flake in the same run, filing it
+                # as infra and dropping it from the rate. A non-firestore health failure
+                # now falls through to the final `else`, where it is still excluded but is
+                # recorded in probe-unmatched.txt instead of being silently miscounted.
+                #
+                # This must precede the hang check either way: a health check fails BY
+                # timing out, so it would otherwise read as the #776 120s hang.
                 infra=$((infra + 1))
-                echo "run $i: INFRA FAILURE (rc=$rc), emulator health check failed, excluded from the rate"
+                echo "run $i: INFRA FAILURE (rc=$rc), firestore emulator health check failed, excluded from the rate"
                 tail -20 "$log"
-              elif grep -q "expected 'loading' to deeply equal 'success'" "$log"; then
-                # The #776 signature specifically, rather than "the job went red".
+              elif grep -q "expected 'loading' to deeply equal 'success'" <<< "$fs_fails"; then
+                # The #776 signature, in firestore's output specifically.
                 flake=$((flake + 1))
                 # Whether the gRPC desync and the #776 assertion co-occur is the whole
                 # question, so count the overlap rather than two independent totals.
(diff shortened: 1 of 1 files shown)


armando-navarro (approved):
Thanks for turning this around so quickly. A clean sweep was not the result you were hoping for, and widening the probe to the workload that actually fails is the right call.

Two things I would fix before anyone runs this for real. The first is one line.

### The probe cannot record a failing run

You did not introduce this, but this PR is where it starts to matter, because the new default mode is the one that is meant to fail sometimes.

The step does not say which shell to use, so GitHub uses its default, which is bash with exit-on-error turned on. The `set -uo pipefail` at the top of the script adds two other flags and does not turn that one off. The loop is:

```
npx firebase emulators:exec ... > "$log" 2>&1
rc=$?
```

With exit-on-error on, a command that fails ends the whole step right there. The next line, the one that reads the exit code, never runs. So the code that decides whether a run was a flake, a hang or an infrastructure problem never gets to see a failing run.

It gets worse after that. The line that writes the tally for an arm sits below the loop, so that arm records nothing. If it is the baseline arm, `probe-counts.tsv` is still empty and the summarize step prints "No arm completed; nothing to summarize." If baseline had already finished, you get baseline's table and silence where override should be.

The result is that the probe can only ever report a clean table. The first time it genuinely reproduces the flake, the job dies and tells you nothing, which your own comment says should mean the probe broke rather than that the flake reproduced. The 120 out of 120 run never hit this, because nothing in it failed.

I checked that `emulators:exec` does exit non-zero on a failed test: a forced #776 failure, an unreachable emulator and missing functions dependencies each gave 1, against 0 for a clean suite. Then I reproduced the abort with the same loop structure, under `bash -e` and under the fuller form GitHub uses for an explicit `shell: bash`. Both behave the same. Run 1 prints, run 2 kills the step, and nothing after the loop executes.

The version of #780 I reviewed had `set +e` and `set -e` wrapped around that command. They came out somewhere before it merged, and I approved that PR without noticing, so this one is at least half mine. Putting them back fixes it, and so would `|| true`, or `if ! npx firebase ...; then rc=$?; else rc=0; fi`.

You wrote that inspection would not have found the health-check bug, that it only showed up against a real log. Same lesson here, pointed at both of us: this survived your authoring, my review of #780, and a second read today, and only fell over when I ran it. It is also why I would want both fixed before the first real run. Merge, run, learn has served you well, and this is the one bug that breaks that loop, because the first run to reproduce a flake is the one that reports nothing.

### The counting still assumes only the firestore tests ran

This one does come from the widening. When the probe ran a single test file, anything it found in the log had to be about #776. That is no longer true, and the searches did not change.

**The flake count picks up other files.** The string it looks for, `expected 'loading' to deeply equal 'success'`, is simply what vitest prints when `expect(result.current.status).toEqual('success')` fails. That is how every data-hook test in this repo is written. It appears in 6 of the 9 test files, 40 times over: firestore 19, database 8, auth 6, useObservable 4, functions 2, storage 1.

I checked this rather than assuming it. I moved `functions/node_modules` out of the way and ran the full suite. Two tests in `functions.test.tsx` failed with that exact message, and your classifier called the run a flake. Your new install step means that particular cause will not come up in the probe, but the general problem stays. The storage upload-timing test, or any non-firestore emulator being slow, would produce the same line.

**The hang count has the same problem.** It searches the whole log for `Test timed out` or `Hook timed out`, and the summary presents that number as the 120 second hang from #776. Any test in any file that times out lands there. Worth noting that `probe-unmatched.txt` is no help here, since it is only written in the final `else`. Anything that lands in flake or hang for the wrong reason does so silently.

**And a real flake can be thrown away.** The health check runs in three files backed by three separate emulators, and your new branch sits above the flake branch. So if one emulator fails its health check while firestore genuinely flakes in the same run, the log contains both messages, the health check wins, and the flake is filed as infrastructure. It does not even reach `probe-unmatched.txt`, since that only happens in the final `else`. Your own local run was that shape: the failure you used to find this bug was the RTDB emulator being unreachable while firestore was fine.

All three go away if the searches are scoped to firestore's output.

My first attempt at that hardcoded the two test names #776 lists, and it was wrong. It would drop a genuine firestore flake that happened to land in a different test in the same file, which is a worse failure than the one it fixes. Scoping to the file avoids both. In a real log the `FAIL` line naming the file and the assertion line sit next to each other, so:

```
grep -A1 -E "FAIL.*test/firestore\.test\.tsx" "$log" | grep -q "expected 'loading' to deeply equal 'success'"
```

I ran that against six logs. It matches the real #776 failure from #781's overnight run, the one I forced by hand, and a firestore failure in a different test. It does not match the functions failure, an unreachable emulator, or a clean run.

One caveat if you use that shape as written. It is a pipeline inside a step that sets `pipefail`, so a SIGPIPE on the first `grep` would return 141 and read as no match. I could only trigger it with tens of thousands of matching lines, and that file has 15 tests, so it does not arise. Take whatever shape you prefer, the ask is the file scope.

The same idea works for the hang search, and for the health check, where the question should be whether the firestore health check failed rather than any of them.

### Smaller: the 180 minute headroom is thinner than a flat 23 seconds suggests

You sized the raise on 23 seconds an iteration, and 400 times 23 is the 153 minutes in your table. Across the last four CI runs, though, the `Run tests` step lands anywhere from 20 to 31 seconds. At 31 it would want about 207 minutes, past the 180.

My local runs do show a warm-up effect, 19 seconds for the first iteration against 15 or 16 after, but I would not lean on it: every iteration still boots five fresh JVMs, and what a timeout cares about is the slowest iterations rather than the median. So I am not claiming the real figure is lower, only that a flat 23 is optimistic. The default is fine under any reading. It is the 200 cap, the setting the raise was sized for, that sits close enough to the wall to be worth another look. And if #776's 120 second hang ever reproduces, those iterations cost far more than 23 seconds each, which is the case where the cap and the wall meet.

One related thing, mostly reassurance. Your comment says the separate summarize step means a job killed by `timeout-minutes` still reports the arms that finished. GitHub's workflow cancellation reference does describe a path where that works: on cancellation the server re-evaluates `if` conditions, `always()` comes out true and the step runs, and there is a five minute window before anything still running is forcibly terminated. What the docs do not say is whether a `timeout-minutes` kill goes down that cancellation path. If it does, you are fine with room to spare, since summarizing takes seconds. Worth knowing only because the tallies are written one per arm, so it is the difference between losing an arm and losing the run.

### What I checked that came out fine

- The full suite runs clean under your new command: 9 files, 69 tests passing, five times out of five, 15 or 16 seconds each on my machine.
- Your timing diagnosis is right. Taking the Node 22 job of #781's overnight run as an example, the `Run tests` step was 23 seconds while the job record was 59. Different run from whichever you measured, so do not read anything into 59 against your 57, but the shape is exactly what you describe.
- Your health check fix works on a log you have never seen. I still had one from Tuesday with the firestore emulator genuinely unreachable: `main` files it as a hang, yours as infrastructure. The `FAIL` anchor earns its place too, since that log has 8 lines carrying the phrase and only 1 with FAIL.
- And it does not misfile the real thing. On the #776 failure in #781's overnight run the health check passed and nothing matches the new pattern, so that run still counts as a flake. One of those passes took 2245ms though, which is worth a look on its own given the failing tests work to a 1000ms budget.

Approving rather than blocking, because merging this does not run anything, and the shell bug is sitting on `main` whether or not this lands. I would just not trust a table produced before both of those are in.

tyler-reitz:
All three are in as `56e354d`. Taking the shell bug first, because the attribution is wrong in my favour.

## The `set -e` abort is mine, not half yours

I removed that `set +e` / `set -e` pair in `43f39e5`, and I wrote in the commit message that it "turned errexit on partway through a script that never had it enabled". The first half is right and the second half is the mistake: the script never enables errexit, but GitHub runs an undeclared `run:` step as `bash -e {0}`, so it is already on and `set -uo pipefail` does not clear it. The pair was load-bearing and I read it as dead weight.

Restored, with a comment saying why it exists, since the reasoning that removed it was superficially sound and would be again.

Verified the way you did rather than by argument. Under `bash -e`, without the guard, **not even the first iteration prints**: the step dies before `rc=$?`. With it, both arms tally with a failing iteration in each and `probe-counts.tsv` gets both rows.

Your framing of the consequence is the part worth keeping: the probe could only ever report a clean table, and the first run to genuinely reproduce the flake is the one that reports least. That is the exact inversion of merge-run-learn, which is why it was worth catching before the first real run rather than after.

## The counting was still assuming one test file

You are right that this comes from the widening, and all three miscounts reproduce. I built a 7 log corpus, one real (#781's overnight #776 failure) and six synthetic, and ran both the old and new logic over it:

| Log | on `main` | scoped | wanted |
| --- | --- | --- | --- |
| real #776 failure (#781 overnight) | FLAKE | FLAKE | FLAKE |
| functions failure | **FLAKE** | INFRA | INFRA |
| RTDB health fails + firestore genuinely flakes | **INFRA** | FLAKE | FLAKE |
| firestore health fails | INFRA | INFRA | INFRA |
| storage test times out | **HANG** | INFRA | INFRA |
| firestore hangs | HANG | HANG | HANG |
| clean run | PASS | PASS | PASS |

Three miscounts fixed, no regressions, and the real failure still classifies correctly.

I took the file scope rather than the test names, for the reason you gave: dropping a genuine firestore flake that lands in a different test is a worse failure than the one it fixes. Confirmed `-A1` is the right window against the real log, where the `FAIL` line naming the file and the assertion sit on consecutive lines.

On the SIGPIPE caveat: I avoided the pipeline entirely rather than relying on it not arising. The `-A1` result goes into a variable and the second search is a here-string, so there is no producer to signal.

The health check is firestore-specific now, so a non-firestore health failure falls through to the final branch. I also widened that branch's capture from an assertion line to the first `FAIL` line, so a genuine failure in another test file shows up in `probe-unmatched.txt` instead of being invisible. Your point that anything landing in flake or hang for the wrong reason did so silently was the one that made me widen it.

## Timeout

Raised to 240 rather than defending 180. You are right that a flat 23s is optimistic and that a timeout cares about the tail: at 31s the 200 cap wants about 207 minutes. The comment now records the 20 to 31 range and says the ceiling exists for the cap, not for the default, so nobody re-derives it from an average later.

I did not chase whether a `timeout-minutes` kill takes the cancellation path that would let the `always()` summarize step run. Worth knowing and not worth blocking on, since it is the difference between losing an arm and losing the run, and the fix if it turns out badly is to write the tally per iteration rather than per arm.

## Not done

The 2245ms health-check pass you spotted. That is a real lead on #776 given the failing tests work to a 1000ms budget, and it deserves its own look rather than a tired Friday one. Noted for next week.

And separately: this is the second time this week your habit of running the thing rather than reading it caught something two passes of review did not. I have started copying it, which is how the health-check bug surfaced, but I clearly did not apply it to my own edit in #780.

